PR #68, opened 2014-09-10
Add filter to only apply redirects on 404.

### LINKED ISSUE #45: Only apply redirects on 404
It would be neat to be able to enable, via a filter, the ability to only apply redirects if the current page is 404ed. This would be useful when SRM is being used to handle redirects from old post slugs to new post slugs.

pdclark:
Resolves a conflict when SRM is being used to handle redirects from old post slugs to new post slugs.

To enable the feature, use:

``` php
add_filter( 'srm_redirect_only_on_404', '__return_true' );
```

See PR #67 for solution without a filter.

Resolves #45.


sxalexander:
@tlovett1 This PR seems to address the issue in exactly the way you recommend.  Can this be reconsidered?  This would address the issue elegantly using an optional filter.

tlovett1:
Open to this if someone wants to reformat to remove conflicts.

grappler:
@tlovett1 I have created an updated PR #144. I hope it is OK.

PR #72, opened 2014-10-19
Implemented ability to import redirects using visual importer

eugene-manuilov:
1. Go to Tools > Import page and click on "Safe Reidrect Manager (CSV)" importer link. 
1. Select your CSV file with redirects, enter delimiter and enclosure characters and submit the form.
1. Select columns mapping and submit the form.
1. All is done, the amount of imported redirects will be shown.

### DIFF
--- a/inc/class-srm-importer.php
+++ b/inc/class-srm-importer.php
@@ -1,0 +1,292 @@
+<?php
+
+class SRM_Importer {
+
+	/**
+	 * Uploaded file ID.
+	 *
+	 * @since 1.7.6
+	 *
+	 * @access private
+	 * @var int
+	 */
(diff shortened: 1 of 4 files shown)


szepeviktor:
Yes, I like SRM because is consumes only a few CPU cycles. It is much like a non-popularish entrerprise plugin.

szepeviktor:
I think popular plugins are full of features for non-experts and those features demand faster processors.

szepeviktor:
@eugene-manuilov Please consider writing a wp-cli command also for CLI guys.
https://github.com/wp-cli/wp-cli/wiki/Commands-Cookbook

PR #75, opened 2014-11-13
Added an extra validation to check if the redirect destination is valid.

vaurdan:
This will validate the $redirect_to to check if it's not empty. Empty $redirect_to causes the page to exit without redirecting. With this validation, if a invalid redirect is found, we just skip it.

This happened while having a redirection without any field filled (no redirect to and no redirect from). The homepage was exiting to a blank page.


vaurdan:
Sorry for the time it took fixing it, I just committed the changes with the correct spacing.

PR #76, opened 2015-01-28
Use home_url instead of site_url for replacement

swalkinshaw:
A common setup is to have your `WP_SITEURL` contain `/wp` but your `WP_HOME` without it (typical sub-directory install). Right now using `site_url()` causes any requested path starting with `/wp` to 404. This is safe to do since the redirects don't run in admin anyway.


tlovett1:
This makes sense. Thank you!

retlehs:
any chance we could get a version bump? or at least get these changes on trunk in the .org svn repo?

PR #90, opened 2015-05-05
Allow for a custom default redirect status

danielbachhuber:


danielbachhuber:
I've also submitted as a patch to VIP.

tlovett1:
Can you reopen against develop?

danielbachhuber:
> Can you reopen against develop?

#91

PR #109, opened 2016-03-13
Add a "notes" field to redirect rules

stevegrunwell:
Similar to the motivation behind #107, sometimes it's helpful to have a textarea where we can put notes about *why* a redirect is in place. Is this a workaround for something? Can we remove it after a certain date?

Along with this new field (which saves to the post_excerpt field), this PR rewrites the meta box to use WP-Admin's `.form-table` pattern for better responsiveness across various screen sizes and devices. Late-escaping has been improved, and the restructuring also makes the interface a little cleaner:

**Before:**
Before

**After:**
After


tlovett1:
Awesome!

PR #116, opened 2017-01-09
Adds a custom capability

fatjester:
Adds a custom capability (instead of using "manage_options") and assigns it to the Administrator Role (the only default wordpress role that has the "manage_options" capability).  This allows role manager plugins to add the capability of "manage_safe_redirects" to other default and custom wordpress roles without giving those roles other features assigned to the "manage_options" role (such as "Tools > Delete Site" in multisite installations)


tlovett1:
Can you fix the formatting?

fatjester:
No Problem.
The formatting has been fixed.

PR #118, opened 2017-01-27
.org Parity

davisshaver:
Demonstrates diff between `wordpress.org` version of plugin.

@tlovett1 what do think of bringing this back up to parity?


davisshaver:
Also updates test matrix per https://github.com/tlovett1/Safe-Redirect-Manager/pull/108/files

PR #137, opened 2017-09-19
Use menu_order to set order of redirect processing

tauno:
Uses the menu_order value to sort the redirects to give more control over the order they are processed in.

Adds ability to import the menu order when creating redirects from a file via wp-cli.


inc/classes/class-srm-post-type.php
```diff
@@ -360,6 +364,7 @@ public function action_custom_redirect_columns( $column, $post_id ) {
 	public function filter_redirect_columns( $columns ) {
 		$columns['srm_redirect_rule_to']          = esc_html__( 'Redirect To', 'safe-redirect-manager' );
 		$columns['srm_redirect_rule_status_code'] = esc_html__( 'HTTP Status Code', 'safe-redirect-manager' );
+		$columns[ 'menu_order'] 									= esc_html__( 'Order', 'safe-redirect-manager' );
```
grappler: Should use spaces and not tabs for indentation.

PR #139, opened 2017-11-01
Fix Home Page Redirects in Subdirectory Installs

### LINKED ISSUE #135: How does it redirect the homepage ?
I thought if I just write "/" in the 'Redirect From' field it would redirect the homepage. It does not seems to be doing that.

Please guide.

ninnypants:
The base site path replacement that happens when prepping the requested path for comparison against redirects creates a `$requested_path` that is an empty string, so you can never match it with a redirect rule (empty rules are skipped and it doesn't match `/`). This moves the check for an empty `$requested_path` so that it's after the replacement fixing this issue.

This will likely fix #135


tlovett1:
Well done!

Issue #140, opened 2017-11-22
Quick Edit removed on all posts


jameswburke:
[The quick edit removal code](https://github.com/tlovett1/safe-redirect-manager/blob/master/safe-redirect-manager.php#L86-L97) targets all post types, not just Safe Redirect Manger. Causes the quick edit button to be removed in all locations.

I have a PR I can submit to resolve this.

tlovett1:
Fixed in 1.8

PR #141, opened 2017-11-22
Fix broken quick edit links

ninnypants:
Currently the removal of quick edit links is too broad, and removes the links from all areas of the site. This makes sure that the removal is limited to just the `redirect_rule` post type.


ninnypants:
This would fix #140

PR #143, opened 2017-11-23
Only disable quick edit for SRM post type

### LINKED ISSUE #140: Quick Edit removed on all posts

### LINKED ISSUE #142: Re-enable edit bulk action
I'd like to advocate for re-enabling the edit bulk action. It is very handy in cases where you need to launch a number of redirects at a time in the future that isn't specific enough for scheduling. You can create drafts of all of them before hand then just switch them on when needed.

jameswburke:
Addresses issues in #140 and #142.

Alternative approach to PR #141.


tlovett1:
Thanks!

PR #144, opened 2017-12-11
Add filter to only apply redirects on 404

### LINKED ISSUE #45: Only apply redirects on 404

grappler:
An update for #68, fixes #45.

I have just copied the code and checked that there is no PHP error when running the code.

props @pdclark


grappler:
@tlovett1 What do you think?

tlovett1:
Love it

PR #146, opened 2018-02-08
remove query string from request path

mslinnea:
It was brought to my attention that this plugin doesn't strip off the query vars when doing the URL comparison.  This issue came up with regard to utm query parameters, but the issue applies to all query parameters.

I think this is a pretty straightforward approach, but I'm definitely open to discussion.  It's possible that this is the intention here, as I realize that maybe users are creating regex strings that include matches on specific query vars.  If so, then an alternative suggestion would be to add a call to apply_filters() on $requested_path so that we could handle this change in our theme code.

### DIFF
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -57,7 +57,7 @@ class SRM_Redirect {
 		}
 
 		// get requested path and add a / before it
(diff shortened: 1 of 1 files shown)


mslinnea:
I'll fix this problem with how the redirects are being created instead.  After some thought, this seems like a big behavior change that would be sure to disappoint some users.

PR #147, opened 2018-02-09
Try to refactor out the create_function call

tomjn:
`create_function` is deprecated in PHP 7.2, this attempts to replace it with an anonymous function

### DIFF
--- a/inc/classes/class-srm-post-type.php
+++ b/inc/classes/class-srm-post-type.php
@@ -160,7 +160,9 @@ class SRM_Post_Type {
 
 		if ( ! empty( $s ) ) {
 			preg_match_all( '/".*?("|$)|((?<=[\\s",+])|^)[^\\s",+]+/', stripslashes( $s ), $matches );
-			$search_terms = array_map( create_function( '$a', 'return trim( $a, "\\"\'\\n\\r " );' ), $matches[0] );
+			$search_terms = array_map( function( $a ) {
+				return trim( $a, "\\"'\\n\\r " );
+			}, $matches[0] );
 		}
 		return $search_terms;
 	}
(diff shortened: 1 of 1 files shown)


tlovett1:
Hey @tomjn, thanks for creating this! We should probably make it a method in the class to preserve 5.2 compat.

tomjn:
@tlovett1 sorry for the delay on this one, I changed it to a static method, and also did a touch of cleanup

inc/classes/class-srm-post-type.php
```diff
@@ -448,17 +453,17 @@ public function action_register_post_types() {
...
+		foreach ( $roles as $role ) {
 			$role = get_role( $role );
 
 			if ( empty( $role ) || $role->has_cap( $redirect_capability ) ) {
 				continue;
 			}
 
-            $role->add_cap( $redirect_capability );
-        }
+            	$role->add_cap( $redirect_capability );
```
grappler: Might as well replace the spaces with tabs.

tomjn: Done

adamsilverstein:
This looks good to me and agree we should get this in ASAP although ultimately I'd love to have tests for `get_search_terms` to back up the change :)

tomjn:
@adamsilverstein agreed but I have not got the bandwidth to continue add tests, and I don't have the bandwidth to chase this up.

Also can there be a full list of the blockers that prevent this from being merged? I get the impression that I make 2 character changes then wait months for a response to be told they were good but I have to do another minor change. Given the urgency now attached to these changes with the shift to PHP 7.2, I'm keen to expedite this

adamsilverstein:
@tomjn The only blocker is time. Given the urgency of this issue, I'll focus on expediting it.

adamsilverstein:
I tested this with PHP 7.2 by performing a search on the redirect rule screen @ `wp-admin/edit.php?post_type=redirect_rule`. before the fix, I get a deprecated error: 



After the change, search works as expected. I also tested with the change and PHP 5.2, search works as expected.

@tlovett1 +1 to merging this

tlovett1:
Thanks all.

PR #148, opened 2018-02-12
Add filters for request path and redirect path

mslinnea:
Any interest in adding filters in these two places for added flexibility?

My particular use case for these filters is to strip off the query string from the requested path, so that the match is done on the path only.  And then I'm adding the query string back on to the redirected URI, so that we don't lose the tracking information provided by the query vars.  If there's demand for those changes in particular, I could open a new PR for that, although I see that as an optional feature, given that some users may be intentionally matching on query vars.


tlovett1:
Merged!

PR #157, opened 2018-10-30
Fix: Handle empty wildcard requests

### LINKED ISSUE #156: Wildcard doesn't match zero characters
Given the following:

- Redirect From: `foo/*`
- Redirect To: `https://example.com/bar/*`

Results in:

1. `https://example.com/foo/baz` correctly redirects to `https://example.com/bar/baz`. i.e. `*` matches `baz`
2. `https://example.com/foo/` fails to redirect to the expected `https://example.com/bar/` i.e. `*` matches `{empty string}`.

The expected behaviour is that the wildcard could be an empty string.

The unit tests do NOT cover this use case.

The current workaround is to add a second redirect rule without the wildcard, which works:

- Redirect From: `foo/`
- Redirect To: `https://example.com/bar/`

GaryJones:
Comments in the code suggest the behaviour to allow `/one/` to redirect when Redirect From is `/one/*` and   Redirect To is `/gohere/*` was meant to be, though there was no specific unit test.

This PR adds a unit test, and a fix.

The fix undoes some workaround code related to trailing slashes, and adds specific code to still allow the same behaviour.

Fixes #156.

Before:

screenshot 2018-10-30 at 19 28 17

After:

screenshot 2018-10-30 at 19 29 11

### DIFF
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -71,7 +71,7 @@ class SRM_Redirect {
 
 		// get requested path and add a / before it
 		$requested_path = esc_url_raw( apply_filters( 'srm_requested_path', $_SERVER['REQUEST_URI'] ) );
-		$requested_path = untrailingslashit( stripslashes( $requested_path ) );
+		$requested_path = stripslashes( $requested_path );
 
 		/**
 		 * If WordPress resides in a directory that is not the public root, we have to chop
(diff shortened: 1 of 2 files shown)


jeremeylduvall:
This issue came up again. @tlovett1 - any chance you could take a peek at this fix over the next few weeks (recognizing holidays are on the horizon)? Thanks!

inc/classes/class-srm-redirect.php
```diff
@@ -136,6 +136,16 @@ public function maybe_redirect() {
...
+				} elseif (
+					// Handle the case of no trailing slashes on redirect from and redirect to, when
+					// requested URL contains the trailingslash.
+					! $matched_path
+					&& ( strlen ( untrailingslashit( $redirect_to ) )  === strlen( $redirect_to ) )
+					&& ( strlen ( untrailingslashit( $redirect_from ) )  === strlen( $redirect_from ) )
+					&& ( strrpos( $normalized_requested_path, '/' ) === strlen( $normalized_requested_path ) - 1 )
+				) {
+					$redirect_to = untrailingslashit( $redirect_to );
+					$matched_path = true;
```
mslinnea: $matched_path should not be set to true here.  At this point, we have only checked that the lengths of the strings are the same and that there is a slash as the last character of the request path.

In my testing, this causes incorrect redirect matching.  To test this out, set up a redirect with no asterisk in the redirect from path and you will see that the logic here causes all request paths that end with a slash to match that.

GaryJones: Good work @mslinnea.

Confirming that commenting out `srm_create_redirect( '/two/', $redirect_to );` (seemingly unrelated to the `/something-else/` request) makes the unit tests all pass, but adding it back in makes the test fail, therefore proving that my initial logic is flawed.

I'm looking again at the logic, but I'm open to suggestions.

GaryJones:
Thanks to @mslinnea's observation and assistance, I've updaed the logic here so that the workaround portion better targets the use case it was meant to, without catching false positives like my first effort did. If you find more false positives caught, let me know and I'll adjust again.

mslinnea:
@GaryJones I believe I've found a solution to this issue.  Please review https://github.com/10up/safe-redirect-manager/pull/161

adamsilverstein:
closing in favor of the more updated https://github.com/10up/safe-redirect-manager/pull/161 - not sure we even need this?

PR #161, opened 2018-12-30
fix issue with redirect rules ending in '/*' not matching correctly

mslinnea:
- Fixes an issue that occurs when a redirect from URL ends with "/*".  

For example a redirect from "/one/*" should successfully match a path of "/one/" as well as a path that has query params attached to it, for example: "/one/?test=true"

I've added tests for this scenario and more.

### DIFF
--- a/tests/test-core.php
+++ b/tests/test-core.php
@@ -510,4 +510,97 @@ class SRMTestCore extends WP_UnitTestCase {
 		fclose( $tmp_file );
 	}
 
+
+    /**
+     * Test a redirect rule that ends with a trailing slash followed by an asterisk.
+     *
+     * @since 1.9.3
+     */
+    public function testWildcardRedirectWithSlash() {
+        $_SERVER['REQUEST_URI'] = '/one/';
+        $redirected             = false;
+
+        // The destination.
+        $redirect_to            = '/gohere/';
+
+        // Create two redirects for testing.
+        srm_create_redirect( '/one/*', $redirect_to );
+        srm_create_redirect( '/two/', $redirect_to );
+        add_action(
+            'srm_do_redirect',
+            function( $requested_path, $redirected_to, $status_code ) use ( &$redirect_to, &$redirected ) {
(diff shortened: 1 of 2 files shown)


tests/test-core.php
```diff
@@ -510,4 +510,97 @@ public function testFileImport() {
...
+        );
+        SRM_Redirect::factory()->maybe_redirect();
+        $this->assertTrue( $redirected, 'Expected that /one/?test=true would redirect to /gohere/?test=true but instead redirected to ' . $redirect_to );
+    }
+
+    /**
+     * Test a URL that shouldn't redirect.
+     */
+    public function testNoRedirect() {
+        // The destination.
```
GaryJones: Is the `REQUEST_URI` intentionally missing here?

tests/test-core.php
```diff
@@ -510,4 +510,97 @@ public function testFileImport() {
...
+        $redirect_to            = '/gohere/*';
+
+        // Create two redirects for testing.
+        srm_create_redirect( '/one/*', $redirect_to );
+        srm_create_redirect( '/two/', $redirect_to );
+
+        // Set up our variable.
+        $redirected = false;
+        /**
```
GaryJones: No need for DocBlock (`/**`) here.

adamsilverstein:
@mslinnea & @GaryJones - thanks for the PRs. Is this still an issue in the plugin? In my testing I could not reproduce.

I have my redirect set up like this:

image

The redirects work as expected, even with query vars. 

Any other tips to reproduce? Maybe this has been fixed already?

adamsilverstein:
ah, got it - I wasn't testing the correct url - testing with `http://developwordpress.localhost/one?test=1` I see the redirect fails.

This looks good. I'm going to follow up with a cleanup PR and document this change.  Thanks again for the help here @mslinnea & @GaryJones.

adamsilverstein:
Closing in favor of https://github.com/10up/safe-redirect-manager/pull/182 where I continued from this PR.

PR #163, opened 2019-02-26
Allow for escaped values on _redirect_rule_from, useful when importin…

raymondware:
Allow for escaped values on _redirect_rule_from, useful when importing regex.


raymondware:
Noticed I had some lint errors on my last PR, should be good now.

PR #168, opened 2019-05-21
Revert "updates readme"

helen:
Reverts 10up/safe-redirect-manager#155

Have to revert this for now so I can try the GH Action updater, SVN won't let me remove or edit the file for some reason so that's basically guaranteed to fail right now..


helen:
Nevermind, managed to delete README.md on .org finally! Guess SVN permissions took a while to catch up today.

PR #171, opened 2019-05-28
Detect duplicate rules for the same 'redirect from' value

### LINKED ISSUE #165: Detect duplicate rules for the same 'redirect from' value
It would be incredibly helpful to notify the user that they've made a duplicate rule for an exisiting `redirect from` value.

On more than one occasion we have noticed duplicate rules being created for the same `from` value when in reality the user should've edited the original rule but wasn't aware one already existed.

I suggest preventing save if the`redirect from` rule is in use by an exisiting rule and outputting an admin notice explaining what's happned with a link to the original rule for them to edit.

Happy to put together a PR to address this if wanted.

sanketio:
### Description of the Change

This plugin was allowed to add duplicate redirection rules with same `redirect from` value. This PR will detect the original rule and will redirect to the edit screen of that rule and delete the duplicate rule.

### Benefits

Users now edit the existing redirection rule instead of adding multiple rules with the same `redirect from` value. And also they do not have to remember whether they have added this rule or not, this PR will take care of that and if existing entry found, it'll redirect to the original rule and delete duplicate one.

### Verification Process

Tested manually on multiple websites.

### Applicable Issues

fixes #165


inc/classes/class-srm-post-type.php
```diff
@@ -448,9 +462,50 @@ public function action_save_post( $post_id ) {
...
+		}
+
+		$redirect_from = get_post_meta( $post_id, '_redirect_rule_from', true );
+
+		if ( empty( $redirect_from ) ) {
+			return $location;
+		}
+
+		// Get post if exist with same redirection rule.
+		$existing_post_id = $wpdb->get_var( $wpdb->prepare( "SELECT post_id FROM $wpdb->postmeta WHERE meta_key=%s AND meta_value=%s", '_redirect_rule_from', $redirect_from ) );
```
adamsilverstein: can we use a WP_Query here instead of a direct $wp_db query? we can limit to fields=['id']

inc/classes/class-srm-post-type.php
```diff
@@ -448,9 +462,50 @@ public function action_save_post( $post_id ) {
...
+	/**
+	 * Add custom query var to indicate that this is a duplicate redirection rule.
+	 *
+	 * @param string $location The destination URL.
+	 * @param int    $post_id  The post ID.
+	 *
+	 * @return string
+	 */
+	public function add_notice_query_var( $location, $post_id ) {
+
```
adamsilverstein: return early if post id not passed or invalid

inc/classes/class-srm-post-type.php
```diff
@@ -252,6 +252,20 @@ public function action_redirect_chain_alert() {
...
+				?>
+				<div class="notice notice-error is-dismissible">
+					<p>
+						<?php esc_html_e( 'Safe Redirect Manager Error: Duplicate redirection rule found. You can modify below rule instead of adding new one.', 'safe-redirect-manager' ); ?>
+					</p>
+				</div>
+				<?php
+
+				// Delete duplicate post.
+				wp_delete_post( $_GET['duplicate-redirect-rule'], true );
```
adamsilverstein: Before deleting the post on this passed id, we should add a nonce check, capability check (can the current user delete this?) and sanitize the $_GET variable for security. Otherwise this could be abused.

adamsilverstein (changes requested):
Nice start on this @sanketio thanks for the PR! I left some small change requests, let me know if you have any questions.

sanketio:
@adamsilverstein : Added feedbacks, please check!

inc/classes/class-srm-post-type.php
```diff
@@ -448,9 +492,78 @@ public function action_save_post( $post_id ) {
...
+
+		$existing_post_id = $existing_post_ids->posts[0];
+
+		// If it is the same post ID, then bail out.
+		if ( (int) $post_id === (int) $existing_post_id ) {
+			return $location;
+		}
+
+		$location = add_query_arg(
+			array(
```
adamsilverstein: I noticed you used `array()` here and `[]` above. Please use one format - whichever best matches what you see elsewhere in the codebase.

adamsilverstein:
@sanketio Code looks good now (left one small comment), going to give this a test.

adamsilverstein:
@sanketio This worked pretty well in my testing. I have some slight language adjustments I will add as suggestions on the PR. Also, I found one bug, when I hit reload on the page where I am editing the duplicate rule, I get an additional error:

image

inc/classes/class-srm-post-type.php
```diff
@@ -252,6 +252,50 @@ public function action_redirect_chain_alert() {
...
+
+				if ( ! wp_verify_nonce( $_wpnonce, 'delete-duplicate-redirect-rule' ) ) {
+
+					$should_delete = false;
+					$error         = __( 'Security check failed when deleting duplicate redirect rule, please contact your administrator!', 'safe-redirect-manager' );
+
+				} elseif ( ! current_user_can( 'delete_post', $duplicate_redirect_rule ) ) {
+
+					$should_delete = false;
+					$error         = __( 'You don\'t have enough permission to delete duplicate redirect rule, please contact your administrator!', 'safe-redirect-manager' );
```
adamsilverstein:
```suggestion
					$error         = __( 'You don\'t have permission to delete the duplicate redirect rule, please contact your administrator!', 'safe-redirect-manager' );
```

inc/classes/class-srm-post-type.php
```diff
@@ -252,6 +252,50 @@ public function action_redirect_chain_alert() {
...
 			}
+
+			$duplicate_redirect_rule = filter_input( INPUT_GET, 'duplicate-redirect-rule', FILTER_VALIDATE_INT );
+
+			if ( ! empty( $duplicate_redirect_rule ) ) {
+
+				?>
+				<div class="notice notice-error is-dismissible">
+					<p>
+						<?php esc_html_e( 'Safe Redirect Manager Error: Duplicate redirection rule found. You can modify below rule instead of adding new one.', 'safe-redirect-manager' ); ?>
```
adamsilverstein:
```suggestion
						<?php esc_html_e( 'Safe Redirect Manager Error: Cannot create rule - duplicate redirection rule found. You can modify the existing rule below.', 'safe-redirect-manager' ); ?>
```

inc/classes/class-srm-post-type.php
```diff
@@ -252,6 +252,50 @@ public function action_redirect_chain_alert() {
...
+				} elseif ( ! current_user_can( 'delete_post', $duplicate_redirect_rule ) ) {
+
+					$should_delete = false;
+					$error         = __( 'You don\'t have enough permission to delete duplicate redirect rule, please contact your administrator!', 'safe-redirect-manager' );
+				}
+
+				if ( $should_delete ) {
+
+					// Delete duplicate post.
+					wp_delete_post( $duplicate_redirect_rule, true );
```
adamsilverstein: @sanketio Deleting a post where the ID is passed in a callback seems like a security risk vector I'd love to avoid introducing if possible. 

I'm wondering if this approach could be shifted to acting at a different point, maybe before the rule is even saved? The idea would be:

* On the "Add Redirect" screen, the "Publish" and "Save Draft" buttons would start disabled. 
* As the user (paused/debounced/cached) types into the Redirect From field, send an REST Ajax request to and endpoint that checks for existing matching rules.
* If there is no match, the rule can be published/saved - the buttons become enabled.
* If there is a match, the buttons remain disabled & a warning message appears with a link to "Edit existing rule" (returned by the endpoint if needed).

I realize this adds some code overhead, however if feels like this would result in both a better user experience and a less risky endpoint (no more arbitrary post deleting!) - what do you think?

One more thing: do we want to try to account for wildcards at all? what happens if I have an existing rule for `/foo/*` and I try to add a rule for `foo/bar/` - does that work correctly?

adamsilverstein:
@sanketio I love where you are headed in this PR, however the deleting post part got me concerned and I wonder if we can refactor a bit to prevent saving entirely if the rule already exists? See https://github.com/10up/safe-redirect-manager/pull/171#discussion_r336584857

jeffpaul:
@sanketio any chance you can respond to the feedback on this PR?

sanketio:
@jeffpaul : Sorry :(, I haven't had the time to check this, but as soon as I get some time, I'll definitely try to take over and modify the way as Adam has suggested.

inc/classes/class-srm-post-type.php
```diff
@@ -48,6 +48,8 @@ public function setup() {
 		add_action( 'admin_print_styles-post-new.php', array( $this, 'action_print_logo_css' ), 10, 1 );
 		add_filter( 'post_type_link', array( $this, 'filter_post_type_link' ), 10, 2 );
 		add_filter( 'default_hidden_columns', array( $this, 'filter_hidden_columns' ), 10, 1 );
+		add_action( 'admin_enqueue_scripts', array( $this, 'load_resources' ) );
```
lukaspawlik: @dhanendran As we're not expecting to get any callback parameters in `load_resources` let's change this as follows

```suggestion
		add_action( 'admin_enqueue_scripts', array( $this, 'load_resources' ), 10, 0 );
```

inc/classes/class-srm-post-type.php
```diff
@@ -48,6 +48,8 @@ public function setup() {
...
+		add_action( 'wp_ajax_srm_validate_from_url', array( $this, 'srm_validate_from_url' ) );
```
lukaspawlik: @dhanendran As we're not expecting to get any callback parameters in `load_resources` let's change this as follows

```suggestion
		add_action( 'wp_ajax_srm_validate_from_url', array( $this, 'srm_validate_from_url' ), 10, 0 );
```

inc/classes/class-srm-post-type.php
```diff
@@ -638,4 +641,54 @@ public static function factory() {
...
 	}
+
+	/**
+	 * Load scripts.
+	 *
+	 * @return void
+	 */
+	public function load_resources() {
+		if ( 'redirect_rule' === get_post_type() ) {
+			wp_enqueue_script( 'redirectjs', plugin_dir_url( 'safe-redirect-manager/safe-redirect-manager.php' ) . 'assets/js/redirect.js', [ 'jquery' ], SRM_VERSION );
```
lukaspawlik:
```suggestion
			wp_enqueue_script( 'redirectjs', plugin_dir_url( 'safe-redirect-manager/safe-redirect-manager.php' ) . 'assets/js/redirect.js', array( 'jquery' ), SRM_VERSION );
```

inc/classes/class-srm-post-type.php
```diff
@@ -638,4 +641,54 @@ public static function factory() {
...
+	 */
+	public function srm_validate_from_url() {
+		$_wpnonce = filter_input( INPUT_GET, '_wpnonce', FILTER_SANITIZE_STRING );
+		if ( ! wp_verify_nonce( $_wpnonce, 'srm-save-redirect-meta' ) ) {
+			echo 0;
+			die();
+		}
+
+		$from = filter_input( INPUT_GET, 'from', FILTER_SANITIZE_STRING );
+		$existing_post_ids = new WP_Query(
```
lukaspawlik, line 685: @dhanendran as we just need to check whether particular post exists, we don't need to populate its meta & term caches so let's ass `update_post_meta_cache` and `update_post_term_cache` to false.

assets/js/redirect.js
```diff
@@ -0,0 +1,27 @@
...
+					action: 'srm_validate_from_url',
+					from: fromRule.val(),
+					_wpnonce: $('#srm_redirect_nonce').val()
+				}
+			).done(function( data ) {
+				if ( '1' === data ) {
+					$('#message').html( '' ).hide();
+					publishBtn.prop('disabled', false);
+				} else if ( '0' === data ) {
+					$('#message').html( '<p>There are some issues validating the URL. Please try again.</p>' ).show();
```
lukaspawlik: @dhanendran Let's use localized strings exposed via [wp_localize_script](https://developer.wordpress.org/reference/functions/wp_localize_script/) instead of hardcoding these here and without any translations.

lukaspawlik (changes requested):
@dhanendran Thanks for your changes! Overall looks good, however I've left you some suggestions and comments.

dhanendran:
@lukaspawlik I have addressed all your comments. Please review again.

lukaspawlik (dismissed):
@dhanendran this looks good to me, over to you for next steps!

jeffpaul:
@Sidsector9 did you want to review this or are you good to merge here?

Sidsector9 (approved):
Found a minor bug which is fixed in [cca0c1c](https://github.com/10up/safe-redirect-manager/pull/171/commits/cca0c1ca6ab90a572c5e28e555a95ef34550882d), looks good otherwise 🚀

PR #176, opened 2019-07-15
Move matching functionality into a separate method

ostowe:
### Description of the Change
I've moved all functionality related to matching a request URL with a redirect into its own method in the `SRM_Redirect` singleton. This method will now return an associative array containing keys for `requested_path`, `redirect_from`, `redirect_to`, and `status_code`, allowing developers to access those values without actually following the redirect.

### Alternate Designs
None considered.

### Benefits
This PR does not change any of the underlying functionality of safe redirect manager, but it does allow more flexibility when integrating with SRM. In my case, I am building a headless WordPress site and need the ability to find a redirect match without actually following the redirect, then pass the resulting information about the match back through WP API.

### Possible Drawbacks
This PR reorganizes code fairly substantially, and may cause merge conflicts with existing PRs if merged.

### Verification Process
* Verified that all existing tests pass.
* Hand-tested a variety of redirects.

### Checklist:
- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.

No new tests are necessary, I don't think.


inc/classes/class-srm-redirect.php
```diff
@@ -165,25 +174,41 @@ public function maybe_redirect() {
...
+	 * Check current url against redirects
+	 *
+	 * @since 1.8
+	 */
+	public function maybe_redirect() {
+		$redirect_match = $this->get_redirect_match();
 
-				header( 'X-Safe-Redirect-Manager: true' );
+		if ( ! empty( $redirect_match['redirect_to'] ) ) {
+			do_action( 'srm_do_redirect', $redirect_match['requested_path'], $redirect_match['redirect_to'], $redirect_match['status_code'] );
```
adamsilverstein: after this change the action only fires if redirect_to is set, before this change it looks like it fired regardless - any reason to change that @ostowe?

inc/classes/class-srm-redirect.php
```diff
@@ -165,25 +174,41 @@ public function maybe_redirect() {
...
+	/**
+	 * Check current url against redirects
+	 *
+	 * @since 1.8
+	 */
+	public function maybe_redirect() {
+		$redirect_match = $this->get_redirect_match();
 
-				header( 'X-Safe-Redirect-Manager: true' );
+		if ( ! empty( $redirect_match['redirect_to'] ) ) {
```
adamsilverstein: maybe this should be checking redirect_from, as originally?

inc/classes/class-srm-redirect.php
```diff
@@ -165,25 +174,41 @@ public function maybe_redirect() {
...
-				do_action( 'srm_do_redirect', $requested_path, $sanitized_redirect_to, $status_code );
+		return $match;
+	}
 
-				if ( defined( 'PHPUNIT_SRM_TESTSUITE' ) && PHPUNIT_SRM_TESTSUITE ) {
-					// Don't actually redirect if we are testing
-					return;
-				}
+	/**
+	 * Check current url against redirects
```
adamsilverstein:
```suggestion
	 * Check current url against redirects.
```

adamsilverstein (changes requested):
@ostowe This is looking good! I left a few small change requests. This could use more testing and ideally unit tests to back up the change.

ostowe:
Thanks for the feedback, @adamsilverstein! I'll review and address as soon as I can.

nicholasio:
Just created another PR to address the same thing: https://github.com/10up/safe-redirect-manager/pull/198

jeffpaul:
Closing in favor of the merged #198.

PR #196, opened 2019-11-22
Fix/189 ignore query params

### LINKED ISSUE #189: Add option to ignore query params
**Describe the bug**
On sites with pretty permalinks enabled, query strings are almost never considered a proper part of the URL, and would be expected to be ignored when looking for redirect matches.

**Steps to Reproduce**
Create a redirect from `/page1/` to `/page2/`.

On navigating to `/page1/` the plugin redirects correctly.
On navigating to `/page1/?param` the plugin does not redirect.

This is causing us significant issues because users are hitting our page to be redirected from Adword links and social media, where various marketting paramaters are included (gclid, utm_source etc). In many cases an option would be to alter the redirect to `/page1/*`, but this doesn't work in cases where the child pages are to stay accessible. (i.e. where you want to redirect `/page1/` without affecting `/page1/child`.

**Expected behavior**
I would expect `/page1/?param` to redirect to `/page2/`.

(It may be necessary to make this behaviour optional to handle sites which are not using pretty permalinks).

**Workaround**

The following regular expression can be used ("Enable Regular Expressions" must be ticked)

`^\/page1(\/)?(\?.*)?$`

**Additional context**

I can see a few ways of resolving this:
- Add a checkbox to the UI labelled something like "Ignore query strings".
- Add a filter to the `maybe_redirect` function to enable ignore query string functionality
- Add a filter allowing a blacklist of query strings to be ignored to be created.

If the query strings are ignored they should be passed on to the redirect-to path, in order to maintain analytics attribution.

The second option above would be my preference. Is this something which is likely to be fixed? If necessary I can contribute a pull request, but it would be nice to have an indication of the preferred way to handle this before I start.

braders:
### Description of the Change

This fixes #189. 

In short, if the redirect is simple (i.e. not regex) and the redirect_from field does not contain query parameters, then any query parameters will be ignored when doing the matching. Any ignored query parameters will be appended to the eventual $redirect_to, to maintain marketing params.

### Possible Drawbacks

This is a possible breaking change, but I consider it unlikely that anyone would be relying on the previous behaviour. Users who want to keep the existing behaviour can use the new `srm_match_query_params` filter

### Verification Process

 - Manually tested against the codebase where the issue was discovered,
 - Tested through `phpunit` tests.

### Checklist:

- [x] All new and existing tests passed.

### Applicable Issues

#189


jeffpaul:
@braders thanks for the PR, I'm pinging @tlovett1 for a review.  Note that we just shipped a release here this past week so it may be a bit before we're able to get another one ready to go, but will let you know of any issues of concern here or when it's good to merge... thanks!

inc/classes/class-srm-redirect.php
```diff
@@ -106,6 +106,15 @@ public function maybe_redirect() {
 			$normalized_requested_path = $requested_path;
 		}
 
+		// normalise the request path with and without query strings, for comparison later
+		$requested_query_params = '';
+		$normalized_requested_path_no_query = false;
+		if (strpos($normalized_requested_path, "?")) {
```
dinhtungdu: using `wp_parse_url`/`parse_url` seems to be safer

dinhtungdu (commented):
This PR conflicts with #198, so I think we should come back to this after merging #198.

dinhtungdu (commented):
Also, the coding style check is failing

inc/classes/class-srm-redirect.php
braders: I've updated this to use `wp_parse_url`/`parse_url`. Please take a second look.

braders:
> This PR conflicts with #198, so I think we should come back to this after merging #198.

I don't see any merge conflicts with this code against develop, and #198 is now merged.

> Also, the coding style check is failing

I believe the code now passes all formatting checks.

inc/classes/class-srm-redirect.php
```diff
@@ -98,6 +98,28 @@ public function match_redirect( $requested_path ) {
...
+		} else {
+			$parsed_requested_path = parse_url( $normalized_requested_path );
+		}
+
+		if ( is_array( $parsed_requested_path ) ) {
+			$normalized_requested_path_no_query = '';
+			$requested_query_params             = '';
+
+			if ( ! empty( $parsed_requested_path['path'] ) ) {
+				$normalized_requested_path_no_query = untrailingslashit( stripslashes( $parsed_requested_path['path'] ) );
```
dinhtungdu: `$parsed_requested_path['path']` will be always exist because the `#normalized_requested_path` variable is not empty.

inc/classes/class-srm-redirect.php
```diff
@@ -98,6 +98,28 @@ public function match_redirect( $requested_path ) {
...
+		$requested_query_params             = '';
+		$normalized_requested_path_no_query = false;
+
+		if ( function_exists( 'wp_parse_url' ) ) {
+			$parsed_requested_path = wp_parse_url( $normalized_requested_path );
+		} else {
+			$parsed_requested_path = parse_url( $normalized_requested_path );
+		}
+
+		if ( is_array( $parsed_requested_path ) ) {
```
dinhtungdu: `$parsed_requested_path` is always an array, so this is always true.

inc/classes/class-srm-redirect.php
```diff
@@ -98,6 +98,28 @@ public function match_redirect( $requested_path ) {
...
+			$normalized_requested_path_no_query = '';
+			$requested_query_params             = '';
```
dinhtungdu: Duplicated variables declaration.
IMO, we can safely use `$parsed_requested_path` without declaring 
intermediate variables.

inc/classes/class-srm-redirect.php
```diff
@@ -116,13 +138,18 @@ public function match_redirect( $requested_path ) {
...
 			} else {
 				if ( $case_insensitive ) {
 					$redirect_from = strtolower( $redirect_from );
 				}
 
-				$matched_path = ( $normalized_requested_path === $redirect_from );
+				// only compare query params if the $redirect_from value contains parameters
+				$match_query_params = apply_filters( 'srm_match_query_params', strpos( $redirect_from, '?' ) );
+
+				$to_match     = ( ! $match_query_params && ! empty( $normalized_requested_path_no_query ) ) ? $normalized_requested_path_no_query : $normalized_requested_path;
```
dinhtungdu, line 142: This line can be simplified to
```suggestion
				$to_match     = $match_query_params ? $normalized_requested_path : untrailingslashit( stripslashes( $parsed_requested_path['path'] ) );
```

dinhtungdu (commented):
@braders  It works! I added some comments to improve code readability.

inc/classes/class-srm-redirect.php
braders: This change causes tests to fail, as untrailingslashit( stripslashes( $parsed_requested_path['path'] ) ) will return an empty string for redirects from the site root.

inc/classes/class-srm-redirect.php
braders: Updated.

inc/classes/class-srm-redirect.php
braders: Updated.

inc/classes/class-srm-redirect.php
braders: Updated.

dinhtungdu (dismissed):
Thanks for your great work @braders, LGTM 🎉

braders:
Now that its been a few months I was just wondering if there is any update on when this might be merged? Thanks!

dinhtungdu (approved):

inc/classes/class-srm-redirect.php
davidmondok: This is unfortunately not correct, as our debug log is getting flooded with undefined index warnings: `Undefined index: path in .../plugins/safe-redirect-manager/inc/classes/class-srm-redirect.php on line 121`

We're not yet sure what's wrong exactly, but it's clear that there's an issue here.

PR #218, opened 2020-10-16
Added header for the redirect from Post ID 

jamesmorrison:
### Description of the Change

This PR adds the redirect post ID to the HTTP response headers. Its' purpose is to aid debugging where a redirect is coming from.

Where there are a large number of similar redirects e.g. for "reviews" it can be hard to pinpoint exactly where the active redirect is coming from.

### Alternate Designs

N/A

### Benefits

The addition of the Post ID will help developers see which redirect rule took effect.

### Possible Drawbacks

The ID of the redirect post will be available to anyone looking at the headers. I don't believe this is a security concern since external users won't be able to achieve anything with this.

### Verification Process

I set up a local install with SRM as the only active plugin; I setup a redirect, cleared the transient cache and loaded the URL that would trigger the redirect rule.

I confirmed that the post ID (`5` in this case) shows in the response headers.

### Checklist:

### Applicable Issues

N / A - this doesn't directly relate to an issue here, I just needed to debug where a redirect was coming from.

### Changelog Entry

Added redirect Post ID to response headers where a redirect rule is invoked.

### DIFF
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -209,7 +209,7 @@ class SRM_Redirect {
 		}
 
 		header( 'X-Safe-Redirect-Manager: true' );
-		header( 'X-Safe-Redirect-Post-ID: ' . esc_attr( $matched_redirect['redirect_post_id'] ) );
+		header( 'X-Safe-Redirect-ID: ' . esc_attr( $matched_redirect['redirect_post_id'] ) );
 
 		// if we have a valid status code, then redirect with it
 		if ( in_array( $matched_redirect['status_code'], srm_get_valid_status_codes(), true ) ) {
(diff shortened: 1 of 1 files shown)


dinhtungdu (commented):
@jamesmorrison This is looking good to me 🎉. I just have a small question/concern about the naming.

`redirect_post_id` makes me think that it the post ID which is redirected from, not the id of redirect itself. `redirect_id` and `X-Safe-Redirect-ID` are clearer for me. What do you think?

jamesmorrison:
@dinhtungdu I updated the redirect post ID header name to `X-Safe-Redirect-ID`

dinhtungdu (approved):

dinhtungdu:
@jamesmorrison Thanks for your work. We will get this one merged shortly.

jeffpaul:
@jamesmorrison I'm seeing a Travis failure, are you able to look into that and see if there's a fix needed to address that?

dinhtungdu:
@jeffpaul look like we have a Travis error for `push` build, this is Travis error, it's not related to this PR. The build for `pr` is passed so it's safe to merge this one.

PR #219, opened 2020-11-06
Fix typo

barryceelen:
interpretted = interpreted


jeffpaul:
@barryceelen nice catch, thanks!

PR #231, opened 2021-03-05
Fix required parameter message in PHP 8.0

### LINKED ISSUE #254: Ensure compatibility with PHP 8
**Describe the bug**
[Originally reported via WordPress.org](https://wordpress.org/support/topic/php-8-0-incompatibility-2/):

> I am on Php 8. When I activate the plugin it gives the error:

```
Deprecated: Required parameter $post follows optional parameter $actions in \wp-content\plugins\safe-redirect-manager\inc\classes\class-srm-post-type.php on line 98
```

> Is it possible for you to fix this so it can work good on Php 8?

**Steps to Reproduce**

**Expected behavior**
Ensure no errors/warnings thrown when running via PHP 8.

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

vinkla:
### Description of the Change

This fixes the following message:

```
Deprecated: Required parameter $post follows optional parameter $actions in safe-redirect-manager/inc/classes/class-srm-post-type.php on line 98
```

`$post` can be either `int|WP_Post|null`: https://developer.wordpress.org/reference/functions/get_post_type/

### Alternate Designs

N/A

### Benefits

The message will now longer be visible.

### Possible Drawbacks

N/A

### Verification Process

I tried to update the plugin in a local WordPress application running PHP 8.0.

### Checklist:

### Changelog Entry

Fix required parameter following optional deprecated message in PHP 8

### DIFF
--- a/inc/classes/class-srm-post-type.php
+++ b/inc/classes/class-srm-post-type.php
@@ -91,11 +91,11 @@ class SRM_Post_Type {
 	 * Remove quick edit
 	 *
 	 * @param  array   $actions Array of actions
-	 * @param  WP_Post $post Post object
+	 * @param  int|WP_Post|null $post Post object
 	 * @since  1.8
 	 * @return array
 	 */
-	public function filter_disable_quick_edit( $actions = array(), $post ) {
+	public function filter_disable_quick_edit( $actions = array(), $post = null ) {
 		if ( 'redirect_rule' === get_post_type( $post ) && isset( $actions['inline hide-if-no-js'] ) ) {
 			unset( $actions['inline hide-if-no-js'] );
 			unset( $actions['view'] );
--- a/inc/classes/class-srm-post-type.php
+++ b/inc/classes/class-srm-post-type.php
@@ -90,7 +90,7 @@ class SRM_Post_Type {
 	/**
 	 * Remove quick edit
 	 *
-	 * @param  array   $actions Array of actions
+	 * @param  array            $actions Array of actions
 	 * @param  int|WP_Post|null $post Post object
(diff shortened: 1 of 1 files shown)


dinhtungdu (commented):
Hi @vinkla, thank you for the PR.

Can you please address the phpcs issue?

About the parameters, I think we should not define `$actions` as an optional parameter. It's a hook so the `$actions` will be always defined already. So I suggest removing the default value of `$actions` instead. What do you think?

vinkla:
> Can you please address the phpcs issue?

Didn't notice the error. Should be fixed now.

> About the parameters, I think we should not define `$actions` as an optional parameter. It's a hook so the `$actions` will be always defined already. So I suggest removing the default value of $actions instead. What do you think?

I'm guessing it would require a major release if we remove the default value? Maybe it could be added as a TODO for the next major release?

dinhtungdu:
> I'm guessing it would require a major release if we remove the default value? Maybe it could be added as a TODO for the next major release?

That method is used for the filter only, so it's safe to change the parameter to not optional. Even it's used by themes or other site plugins, it's safe too because the second parameter is not optional, so the first one is always passed already. So I think we can change it now.

cc @tlovett1

vinkla:
I've update the pull request. Even if it is only used internally it could break any other packages extending this plugin.

dinhtungdu (approved):
Thank you @vinkla for your contribution!

PR #239, opened 2021-04-21
Allow response codes 400 and up

### LINKED ISSUE #205: Add 410 Gone Support
**Is your enhancement related to a problem? Please describe.**

_Google **previously** processed 404 Errors and 410 Errors identically. This frequently left many 404s on URLs that were in the crawl list, the index, and listed as 404 Errors in Google Search Console. Google now uses 404s as a factor in website ranking and domain delisting._

Currently, the `410` code has been changed at Google to this behavior:

1. ASAP Remove URL from the crawl list. (`404` does **_not_** do this)
2. ASAP Remove URL from the Index. (`404` does **_not_** do this)
3. Stop crawling the URL (unless a link to it is found again) (`404` continues to crawl the URL for _up to a year_, wasting Google's bandwidth - and yours too!).
4. A `410` for a short time, **_after a 301_** is removed, will prevent `404s `in _Google Search Console_. 

**Describe the solution you'd like**

Add a `410 Gone` Selection in the dropdown with the `404 Not Found`

**Designs**

Copy the code for `404 Not Found` and modify it to `410 Gone`

**Describe alternatives you've considered**

- I was able to do this with the much heavier [Redirection Plugin](https://wordpress.org/plugins/redirection/) on any platform.

- I was able to do this using .htaccess on Apache from [Pair Networks](https://www.pair.com/).

- I am now teaching Newbies to use a Linode - Ubuntu - [WordOps](https://wordops.net/) - Nginx + MariaDB - WordPress + [YOURLS](https://yourls.org/)  - Which is easier, faster, and has fewer problems than the Apache install mentioned. Of course, it also includes the YOURLS URL Shortener!

- I hope to include the _Safe Redirect Manager_ so that Noobs working on both Apache **_and_** Nginx can easily set up `301s` and `410s` with one lightweight plugin.

**Additional context**

**_Two examples of how it should work:_**

**First Assume:** _Google Search Console_ reports a 404 Error that has no links to it and is not used/needed. Such as a test page or a coming event page that has past and was deleted.

- Step 1 - Set up a `410 Gone` for that URL.

- Step 2 - _Google Search Console_ deletes the `404 Not Found` and now reports a `410 Gone`, (within a week or so, or use _Google Search Console_ to recrawl the URL).

- Step 3 - Remove the `410 Gone` for that URL.

**First Results:** Both `404` and `410` are removed from _Google Search Console_.

**Second Assume:** You wish to remove a page from your website.

- Step 1 - Set up a `301` for that URL to a new page with new info.

- Step 2 - Work to remove any links to that page. For a test page, this may take minutes. Or it may take months for a popular page (perhaps you should rebuild and republish popular URLs?)

- Step 3 - Set up a `410 Gone` for that URL.

- Step 4 - The `410 Gone` will appear in your _Google Search Console_.

- Step 5 - Remove the `410 Gone` for that URL.

**Second Results:** When you remove the `301` there is no `404` in your _Google Search Console_.

**Conclusions:**

_This should be useful in an SEO Strategy as many `404` URLs in the crawl list have proven to depreciate a site._

Persons who type in a faulty URL will still receive a 404. For this, I would suggest a custom 404 page, to lead them back into your site.

### LINKED ISSUE #241: `wp_safe_redirect` does not support HTTP response codes 400 and up
As reported in #239, coming out of #205. This issue is to make sure we document the base issue and work through solutions before deciding what direction to go with this plugin.

As of WordPress 5.4, `wp_safe_redirect()` and `wp_redirect()` only allow for `3xx` response codes ([commit](https://github.com/WordPress/wordpress-develop/commit/40946e984a5c6b775a0e34d86b36a004c6a70204)). This does seem correct to me, as other codes are not relevant to redirects. However, in #205 a very understandable use case for managing `4xx` codes (specifically `410`) is outlined:

<hr />
**You wish to remove a page from your website.**

* Step 1 - Set up a 301 for that URL to a new page with new info.
* Step 2 - Work to remove any links to that page. For a test page, this may take minutes. Or it may take months for a popular page (perhaps you should rebuild and republish popular URLs?)
* Step 3 - Set up a 410 Gone for that URL.
* Step 4 - The 410 Gone will appear in your Google Search Console.
* Step 5 - Remove the 410 Gone for that URL.

Second Results: When you remove the 301 there is no 404 in your Google Search Console.
<hr />

What should Safe Redirect Manager do here to support its most common use cases while still remaining, at heart, a redirect manager and not an HTTP code manager?

aosmichenko:
Use wp_safe_redirect, but actually allow response code of 400 and higher. Redirect will not happen.
Related to https://github.com/10up/safe-redirect-manager/issues/205

### Description of the Change

This will send custom HTTP response following code base of wp_safe_redirect. But will actually allow 400+ response codes

### Alternate Designs

no other options since WordPress core is limited to 300-399 response codes.

### Benefits

This will allow main response code of 410 Gone to send signal to google that page were actually removed.

### Possible Drawbacks

Require manual update in case WordPress function wp_safe_redirect() will get updated.

### Verification Process

- How did you verify that all new functionality works as expected?
> tested code on development website

- How did you verify that all changed functionality works as expected?
> tested responses higher than 399

- How did you verify that the change has not introduced any regressions?
It is new

### Checklist:

- [ x ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ x ] My code follows the code style of this project.

### Applicable Issues

https://github.com/10up/safe-redirect-manager/issues/205

### Changelog Entry

add custom http response

### DIFF
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -243,7 +243,61 @@ class SRM_Redirect {
 
 		// if we have a valid status code, then redirect with it
 		if ( in_array( $matched_redirect['status_code'], srm_get_valid_status_codes(), true ) ) {
-			wp_safe_redirect( $matched_redirect['redirect_to'], $matched_redirect['status_code'] );
+			if ( $matched_redirect['status_code'] > 399 ) {
+				$location = $matched_redirect['redirect_to'];
+				$status = $matched_redirect['status_code'];
+				/**
+                 * Filters the redirect location.
+                 *
+                 * @since 2.1.0
+                 *
+                 * @param string $location The path or URL to redirect to.
+                 * @param int    $status   The HTTP response status code to use.
+                 */
+                $location = apply_filters( 'wp_redirect', $location, $status );
+
+                /**
+                 * Filters the redirect HTTP response status code to use.
+                 *
+                 * @since 2.3.0
+                 *
+                 * @param int    $status   The HTTP response status code to use.
+                 * @param string $location The path or URL to redirect to.
+                 */
+                $status = apply_filters( 'wp_redirect_status', $status, $location );
+
+                if ( ! $location ) {
+                        return false;
+                }
+
+
+                $location = wp_sanitize_redirect( $location );
+
+                if ( ! $is_IIS && 'cgi-fcgi' !== PHP_SAPI ) {
+                        status_header( $status ); // This causes problems on IIS and some FastCGI setups.
+                }
+
+                /**
+                 * Filters the X-Redirect-By header.
+                 *
+                 * Allows applications to identify themselves when they're doing a redirect.
+                 *
+                 * @since 5.1.0
+                 *
+                 * @param string $x_redirect_by The application doing the redirect.
+                 * @param int    $status        Status code to use.
+                 * @param string $location      The path to redirect to.
+                 */
+                $x_redirect_by = apply_filters( 'x_redirect_by', $x_redirect_by, $status, $location );
+                if ( is_string( $x_redirect_by ) ) {
+                        header( "X-Redirect-By: $x_redirect_by" );
+                }
+
+                header( "Location: $location", true, $status );
+	
+			} else {
+				wp_safe_redirect( $matched_redirect['redirect_to'], $matched_redirect['status_code'] );	
+			}
 		} else {
 			wp_safe_redirect( $matched_redirect['redirect_to'] );
 		}


okadots:
I incorporated the changes noted above into my local copy of the plugin, but it does not redirect to another page for any of the 4xx statuses - it simply shows the default browser status message.  For pages with a 4xx status, I would expect that if a user went to that page, they should see an error page, similar to how the default 404 behavior is handled in wordpress using the `404.php` template.

I have modified the above code to produce this behavior. This requires further modification to actually include some default `4xx.php` templates in the plugin, but I have written it so that if there isn't a template, you get a simple error message. I am providing a `410.php` template in my theme.

```// if we have a valid status code, then redirect with it
		if ( in_array( $matched_redirect['status_code'], srm_get_valid_status_codes(), true ) ) {
			if ( $matched_redirect['status_code'] > 399 ) {
				$status = $matched_redirect['status_code'];

				if ( ! $is_IIS && 'cgi-fcgi' !== PHP_SAPI ) {
					status_header( $status ); // This causes problems on IIS and some FastCGI setups.
				}
				
				//display an error template based on the status
				if( ! locate_template( $status.'.php', true ) )
					echo 'Sorry, the page you requested has been permanently removed.';

				exit;

			} else {
				wp_safe_redirect( $matched_redirect['redirect_to'], $matched_redirect['status_code'] );	
                                 exit;
			}
		} else {
			wp_safe_redirect( $matched_redirect['redirect_to'] );
		}

dkotter:
@aosmichenko Thanks for the PR! This looks like a good start to me but I think there's some changes we need here. Let me know if you're still interested in working on this or if you'd like me to finish this off.

If someone is using these 4** status codes, we don't actually want to redirect, we want to either output an error message or render a template that can have a message (and ensure we output the right status code in both instances). There some good direction in the comment from @okadots, so I'd recommend looking at that.

Looking at your code, we can get rid of anything that looks at the `$location`, since we aren't going to be redirecting. I think we can also get rid of the filters around the location and status params, since I don't think those will be needed either.

In short, if the status code is above 399, we want to output the right status code and then try loading a template. If that template doesn't exist (open to suggestions on if we should add templates to our plugin or just rely on those templates being added to themes), we should output an error message and ensure that message has a filter around it so it can be easily changed.

jeffpaul:
Closing in favor of #300.

PR #242, opened 2021-05-23
Fixes #145

### LINKED ISSUE #145: Redirect Limits
How can we change the default of 250 redirects in the updated version? The old code hack does not work anymore (https://www.isrgrajan.com/how-to-increase-redirect-limits-in-safe-redirect-manager-wordpress-plugin.html)
Please help!

sultann:
The PR fixes the issue Redirect Limits #145.

- Separate function for redirection limit
- Increased limit to 1000
- Transient cache key changed as per the limit

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -415,5 +415,5 @@ function srm_match_redirect( $path ) {
  * @return int
  */
 function srm_get_max_redirects() {
-	return apply_filters( 'srm_max_redirects', 2 );
+	return apply_filters( 'srm_max_redirects', 1000 );
 }
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -71,7 +71,8 @@ function srm_get_redirects( $args = array(), $hard = false ) {
 
 		}
 
-		set_transient( $transient_key, $redirects );
+		// Set transient to 30 days to remove old transients if the max redirects changes.
+		set_transient( $transient_key, $redirects, 30 * DAY_IN_SECONDS );
 	}
 
 	return $redirects;
@@ -132,9 +133,8 @@ function srm_get_valid_status_codes_data() {
  * @since 1.8
  */
 function srm_flush_cache() {
-	delete_transient( '_srm_redirects' );
+	delete_transient( '_srm_redirects_' . srm_get_max_redirects() );
 }
-
 
 /**
  * Check for potential redirect loops or chains
@@ -415,5 +415,5 @@ function srm_match_redirect( $path ) {
  * @return int
  */
 function srm_get_max_redirects() {
-	return apply_filters( 'srm_max_redirects', 1000 );
+	return apply_filters( 'srm_max_redirects', 2 );
 }
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -14,8 +14,9 @@
  * @return array $redirects An array of redirects
  */
 function srm_get_redirects( $args = array(), $hard = false ) {
-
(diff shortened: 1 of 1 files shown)


inc/functions.php
```diff
@@ -14,8 +14,9 @@
  * @return array $redirects An array of redirects
  */
 function srm_get_redirects( $args = array(), $hard = false ) {
-
-	$redirects = get_transient( '_srm_redirects' );
+	$default_max_redirects = srm_get_max_redirects();
+	$transient_key         = '_srm_redirects_' . $default_max_redirects;
```
dinhtungdu, line 18: Hi @sultann, thanks for the PR! Can you explain why you change the transient key here?

sultann: @dinhtungdu, I changed the transient key because the redirection limit is filterable. If someone filters the redirection limit, he will still get the old saved value. So I added the limit as the key. If the limit is changed, it will return the updated data.

dinhtungdu: @sultann I don't think it's necessary to change the key. The transient gets deleted every time we create/udpate/detete a redirect rule. So it doesn't make sense to me to keep the old value.

sultann: @dinhtungdu, the transient gets updated on any CRUD action that's correct, but if someone changes the redirection limit using filter without doing any crud action, that will still return the old saved value until the transient gets cleared.
Do you still want me to change that part?

dinhtungdu: > but if someone changes the redirection limit using filter without doing any crud action, that will still return the old saved value until the transient gets cleared.

@sultann The transient still holds the old saved value which is all redirects. And that's is correct behavior IMO. What is the problem with the old saved value at this point, can you clarify?

sultann: @dinhtungdu, the main point of adding that dynamic transient key is, say, you have 1200 redirects, and when the function gets called, it will query until the maximum limit 1000(right now we set) reached & set cache for future finally returns redirects. 

Now you realized you are getting only 1000, but you have 1200 redirects, so you changed the limit to 1500 using the filter. But that function will still return 1000 redirects since that's cached, and the function does not care if the maximum limit is changed or not. But when we make the cache key using the maximum limit, and you call the function after changing the maximum redirection limit, it will return the new sets of redirects immediately without clearing the cache. 

If that's still not an acceptable reason, I would better remove that part and update the PR.

dinhtungdu: @sultann thanks for the clarification!

> @dinhtungdu, the main point of adding that dynamic transient key is, say, you have 1200 redirects, and when the function gets called, it will query until the maximum limit 1000(right now we set) reached & set cache for future finally returns redirects.

When you have the redirect limit set to `1000`, you can't create more than 1000 redirects, so you can't have 1200 redirects at that point. We prevent creating redirects when we reach the limit here: https://github.com/10up/safe-redirect-manager/blob/07c35d5ccc913cd82f26c439efa9c25ab571c750/inc/classes/class-srm-post-type.php#L235-L247

But we're only checking the max redirect on the WordPress add new redirect screen, we can still add more redirects than the limit by importing the redirect rules using CLI. So your concern makes sense to me now.

dinhtungdu (approved):
This PR looks good to me now!

In https://github.com/10up/safe-redirect-manager/pull/242/commits/b8f9b51be242192d8fff458a3a0752cdaf1ad359, I fixed the cache flushing issue caused by the new transient name format.

PR #248, opened 2021-07-13
Check non-active multisite directory against the main site redirects

### LINKED ISSUE #212: Document or update multisite subdirectory site behavior
For WordPress multisite use, this plugin runs into an edge case with subdirectory sites:

example.com/my-site

To add a redirect to this site, it needs to added to the subsite directly, and not from the primary example.com domain using the /my-site path, as we might expect. WordPress sends the request to the subsite before checking the primary site, where the rule can work.

If this subsite is archived, marked as spam, or marked as deleted (but not actually deleted), we can't cover it with any redirect rules - rules on the subsite no longer work, but rules on the main site don't work either, because WordPress still sees it as an available multisite and doesn't send the request to the primary domain for redirection.

If the subsite is physically deleted (say, after export to another domain), only then we can use SRM to create redirects on the main site to cover that path.

However, these rules are likely to conflict with Nginx/Apache rewrite rules which support subdirectory subsite usage of /wp-admin, and therefore no SRM rules can rewrite example.com/my-site/wp-admin because it's already been rewritten at the server level before it gets to WordPress. This is problematic for trying to fully redirect a subdirectory subsite to a new domain.

My suggestion would be some multisite documentation pointing out this case, clarifying that
- active subdirectory subsites need their own rules
- deleted subsites must be fully deleted for primary site rules to cover their paths, and to avoid grey-area state where no rules can be applied on the deactivated path
- Higher-level server rules will prevent SRM from fully redirecting moved subdirectory subsites 

It is more obvious to me that subsites with subdomains or mapped domains would get their own rules or need to be managed at a server level if those domains change, but that may also be worth a mention.

phpbits:
### Description of the Change

This PR contains fixes for this issue: https://github.com/10up/safe-redirect-manager/issues/212. Whenever the multisite subsite is not active, check the requested path against the saved redirect rules from the main website.

### Verification Process

- Create a multisite installation with several sub-sites.
- Archive, deactivate or mark a site as spam
- Go to the main website and add a redirect rule for the subsite you've modified the status from step 2.
- Verify that the redirect is working.

### Checklist:


dinhtungdu (commented):
@phpbits I can't get this working. I use a WordPress Subdirectory Multisite with Apache as the webserver, the pretty permalink is turned on. I'm using standard WordPress access rules for multisite. See the screencast: https://share.getcloudapp.com/geugerel

jeffpaul:
With this still in flux and the 1.10.1 release pending, I'm punting this to the 2.0.0 milestone to hopefully get resolved then.

ciprianimike:
the `maybe_multisite()` function is called within `maybe_redirect()`, registered to the `template_redirect` hook.

Given the scenario described in #212 , the code attached to the `template_redirect` hook **will never be executed**.
WordPress interrupts the hook sequence when it checks the blog status, returning the message `This site has been archived or suspended.`

I am refactoring this approach.

dinhtungdu:
I'm so sorry for this super late review! Testing against latest commit, I'm still having the same issue as my previous test. Here are my steps:

1. Create a sub directory multisite.
2. Create a sub site with slug set to `two`.
3. Archive that site.
4. Create a redirect rule in the main site from `/two/test/` to an arbitrary external site.
5. Try visiting `/two/test`, see the Archived site notice instead of the external site set in step above.

@ciprianimike can you please verify these steps on your end?

<details>
<summary>Screenshots</summary>

Screen Shot 2021-11-08 at 11 29 01
Screen Shot 2021-11-08 at 11 30 25
Screen Shot 2021-11-08 at 11 30 36

</details>

gsarig:
@dinhtungdu Did you, by any chance, have the plugin deactivated on the archived site? 

I ran multiple tests on various different environments following the same steps as you, and @ciprianimike's fix seems always to work as expected. The only time it didn't work, was when I had the plugin inactive on the archived site. In that scenario, it makes sense that the redirect is never triggered, as we are still on the archived site and not on the main one. That's why deleting the site instead of archiving it restores the redirection: the second site no longer exists, and the main site's plugin, which is active, executes the code.

jeffpaul:
@Sidsector9 this is waiting on your review

Sidsector9 (approved):
Took me some time to get this working, these are the steps I followed:

1. Create a multisite (Apache Server)
2. Create 3 subdirectory sites - /one, /two, /three
3. Create a /two/about page
4. Network activate SRM
5. Mark /two as spam
6. Add rule to main site - from `/two/about` to `https://www.google.com`
7. Visit `http://mainsite.com/two/about`
8. Verified redirection to `https://www.google.com`

PR #277, opened 2022-06-30
Add support for notes when importing redirects

barryceelen:
Safe Redirect Manager allows adding notes when creating a redirect via its admin settings page. This update adds support for importing notes via the CLI import command and the `srm_import_file()` function, and for creating a note when using the  `srm_create_redirects()` function.


cadic (changes requested):
@barryceelen thank you for this update

Could you please update Unit tests to pass with the new `notes` argument

```
1) SRMTestCore::testFileImport
Undefined index: notes
```

jeffpaul:
@barryceelen are you able to help update the unit tests here?

cadic (approved):
@jayedul thanks for providing the fix to Unit Tests

LGTM

PR #279, opened 2022-08-18
Replace domain when the redirect url is a relative path

### LINKED ISSUE #269: Regex redirects without leading `/` are buggy
### Describe the bug

Enabling a regular expression without a leading `/` can be quite buggy.

* in most cases I am getting redirected to the wp-admin
* if a full URL is added added as the destination, it will include the domain twice.

### Steps to Reproduce

Scenario One

1. Create a regular expression redirect
    - Redirect from: `(go|be)\/h{0,}$`
    - Redirect to: `/404-regex`
2. In the CLI run `curl -I http://xu-osp-plugins.local/be/hhhh`
3. Observe the location header is `http://xu-osp-plugins.local/wp-admin/`

Scenario Two

1. Create a regular expression redirect
    - Redirect from: `(go|be)\/h{0,}$`
    - Redirect to: `http://xu-osp-plugins.local/404-regex`
2. In the CLI run `curl -I http://xu-osp-plugins.local/be/hhhh`
3. Observe the location header is `/http://xu-osp-plugins.local/wp-admin/` (note leading slash)

_No response_

### Environment information

- Device MacBook Pro
- OS: MacOs 11.6.5
- Browser: curl 7.64.1 

### WordPress information

WordPress 6.0
WordPress 5.9
WordPress 5.8

### Code of Conduct

dhanendran:
### Description of the Change
We need to add the leading `/` only when the `redirect_to` value is a relative path.
 
Closes #269 

#### Scenario One

1. Create a regular expression redirect
2. Redirect from: `(go|bee)\/h{0,}$`
3. Redirect to: `/404-regex`
4. In the CLI run `curl -I http://xu-osp-plugins.local/bee/hhhh`
5. Observe the location header is `http://xu-osp-plugins.local/wp-admin/`

#### Scenario Two

1. Create a regular expression redirect
2. Redirect from: `(go|bee)\/h{0,}$`
3. Redirect to: `http://xu-osp-plugins.local/404-regex`
4. In the CLI run `curl -I http://xu-osp-plugins.local/bee/hhhh`
5. Observe the location header is `http://xu-osp-plugins.local/wp-admin/` (no leading slash)

### Changelog Entry
Fix - Regex redirects without leading `/` are buggy

### Credits
Props @dhanendran 

### Checklist:

### DIFF
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -184,7 +184,7 @@ class SRM_Redirect {
 				}
 
 				// Allow for regex replacement in $redirect_to
-				if ( $enable_regex ) {
+				if ( $enable_regex && ! filter_var( $redirect_to, FILTER_VALIDATE_URL ) ) {
 					$redirect_to = preg_replace( '@' . $redirect_from . '@' . $regex_flag, $redirect_to, $requested_path );
 				}
 
--- a/inc/classes/class-srm-redirect.php
+++ b/inc/classes/class-srm-redirect.php
@@ -186,6 +186,7 @@ class SRM_Redirect {
 				// Allow for regex replacement in $redirect_to
 				if ( $enable_regex && ! filter_var( $redirect_to, FILTER_VALIDATE_URL ) ) {
 					$redirect_to = preg_replace( '@' . $redirect_from . '@' . $regex_flag, $redirect_to, $requested_path );
+					$redirect_to = '/' . ltrim( $redirect_to, '/' );
 				}
 
 				// re-add the query params if they've not already been added by the wildcard


peterwilsoncc:
Scenario one still seems buggy to me, I'd expect the redirect to be to `http://xu-osp-plugins.local/404-regex` rather than the WordPress dashboard.

The fix for scenario two is working as expected.

dhanendran:
@peterwilsoncc For some reason SRM is adding 2 leading slashes `//` which then considered as full URL but not a valid one, so WP redirects to `wp-admin` page. I have added a fix and it should work now.

peterwilsoncc (approved):
LGTM and tests well. Thank you.

PR #300, opened 2023-01-18
Feature/handle 4xx status codes

### LINKED ISSUE #241: `wp_safe_redirect` does not support HTTP response codes 400 and up

noraconley:
### Description of the Change

Handles 403, 404, and 410 status codes. 403 and 410 status codes will add a header and die using `wp_die`. 404 status code will render the 404 template. This will happen even in environments that "soft 404" by default (for example, some environments when "Plain" permalinks are set).

Closes #241 

### How to test the Change

- Create a Redirect Rule that has a status code of 403. On the front end, this request should die with a status code of 403
  - Repeat this step for 410
- Create a Redirect Rule that has a status code of 404. This should render the 404 template with a status code of 404.
  - Repeat this test with permalinks set to "Plain"

### Changelog Entry

> Added - Handling of 403, 404, and 410 status codes

### Credits

Props @nateconley, @aosmichenko

### Checklist:


inc/classes/class-srm-redirect.php
```diff
@@ -142,8 +142,8 @@ public function match_redirect( $requested_path ) {
 			$enable_regex = ( isset( $redirect['enable_regex'] ) ) ? $redirect['enable_regex'] : false;
 			$redirect_id  = $redirect['ID'];
 
-			// check if the redirection destination is valid, otherwise just skip it
-			if ( empty( $redirect_to ) ) {
+			// check if the redirection destination is valid, otherwise just skip it (unless this is a 4xx request)
+			if ( empty( $redirect_to ) && 403 !== $status_code && 404 !== $status_code && 410 !== $status_code ) {
```
cadic: Maybe change this to `in_array()` for the better readability?

cadic (changes requested):
@nateconley thank you for the great work in this PR!

The code looks good, I have just one small question/suggestion to change multiple equations to a single in_array().

Tested the feature, works as expected.

Could we add a message to the 403/410 pages describing what happened? Having an empty die-box is confusing. I believe we could use the existing "Redirect Note" field or add a new input for that.

image

inc/classes/class-srm-redirect.php
noraconley: Thank you for the feedback and I definitely think a message would be a great improvement. For the message, I added a new field. If we were to use the "Redirect Notes" field, upon update, users could now have once admin-only meta data now public facing. Adding a new field allows users to opt-in.

I updated the logic to use `in_array` where you mentioned, but I left the logic in place in other places where there are only two comparisons. Let me know if you think changing in the other places would make it more readable!

cadic (dismissed):
LGTM, thanks @nateconley!

Tested 4** pages with messages, works great

jeffpaul:
@dkotter holding on merge here as I'd like your review as well given this is a long-running conversation in the plugin

inc/classes/class-srm-post-type.php
```diff
@@ -582,6 +590,12 @@ public function redirect_rule_metabox( $post ) {
 			<em><?php esc_html_e( "If you don't know what this is, leave it as is.", 'safe-redirect-manager' ); ?></em>
 		</p>
 
+		<p id="srm_redirect_rule_message_container">
+			<label for="srm_redirect_rule_message"><strong><?php esc_html_e( 'Message:', 'safe-redirect-manager' ); ?></strong></label>
+			<textarea name="srm_redirect_rule_message" id="srm_redirect_rule_message" class="widefat"><?php echo esc_attr( $redirect_message ); ?></textarea>
```
dkotter: Seems like we may want to use `esc_textarea` here instead of `esc_attr`, unless that causes some formatting issues?
```suggestion
			<textarea name="srm_redirect_rule_message" id="srm_redirect_rule_message" class="widefat"><?php echo esc_textarea( $redirect_message ); ?></textarea>
```

README.md
```diff
@@ -54,6 +54,8 @@ so you shouldn't be serving stale redirects.
 `srm_max_redirects` that you can utilize to up this number.
 * "Redirect From" and requested paths are case insensitive by default.
 * Developers can use `srm_additional_status_codes` filter to add status codes if needed.
+* Rules set with 403 and 410 status codes are handled by applying the HTTP status code and rendering a blank page.
```
dkotter: Seems like this isn't 100% true anymore. There is an option to render a message instead of just a blank page (would also need updating in readme.txt)

assets/js/redirect.js
```diff
@@ -23,5 +23,43 @@
 				}
 			});
 		});
+
+		// Disable the 'Redirect To:' field unless a 4xx status code is set.
```
dkotter: Seems like this code comment says the opposite of what actually happens:

```suggestion
		// Disable the 'Redirect To:' field if a 4xx status code is set.
```

inc/classes/class-srm-post-type.php
```diff
@@ -570,6 +577,7 @@ public function redirect_rule_metabox( $post ) {
 			<label for="srm_redirect_rule_to"><strong><?php esc_html_e( '* Redirect To:', 'safe-redirect-manager' ); ?></strong></label><br />
 			<input class="widefat" type="text" name="srm_redirect_rule_to" id="srm_redirect_rule_to" value="<?php echo esc_attr( $redirect_to ); ?>" />
 		</p>
+		<p class="description" id="srm_to_disabled_message" style="display:none;"><em><?php esc_html_e( 'No "Redirect to" value applies for 4xx error codes.', 'safe-redirect-manager' ); ?></em></p>
```
dkotter: I'd suggest tweaking this text a bit:

```suggestion
		<p class="description" id="srm_to_disabled_message" style="display:none;"><em><?php esc_html_e( 'The "Redirect to" value doesn\'t apply for 4xx error codes.', 'safe-redirect-manager' ); ?></em></p>
```

inc/classes/class-srm-post-type.php
dkotter: I see the below code is already using `esc_attr` which is why we probably did the same here. I'd suggest testing to see if `esc_textarea` works like we want but if not, we can use `esc_attr` instead

dkotter (changes requested):

dkotter:
@nateconley Thanks for the work here! Tested this out and all works as expected. Left a few minor comments but once those are taken care of, this will be good to merge.

inc/classes/class-srm-post-type.php
Sidsector9: @dkotter I just checked the code and found that we sanitize the `textarea` data using `sanitize_text_field` instead of `sanitize_textarea_field` before saving it to DB. It wouldn't make much of a difference using `esc_attr` vs `esc_textarea` in this situation.

jeffpaul:
Note that there may be some folks who worked in #239 that we should include in the credits/props on this PR.

noraconley:
@dkotter I have pushed updates based on your feedback.

dkotter (approved):

ocean90:
This change will cause a lot of undefined `message` index PHP warnings unless the transient used for caching `srm_get_redirects()` is deleted by `srm_flush_cache()`.

dkotter:
Thanks for pointing that out @ocean90. I have a PR up to address this

PR #301, opened 2023-01-24
Custom protocol per route

### LINKED ISSUE #46: Redirection to HTTPS
It would be super useful to me if this plugin could handle redirecting http to https for certain URL's. Thoughts on this @danielbachhuber ?

jayedul:
### Description of the Change
Custom protocol selection feature per route implemented. 

Closes #46 

### How to test the Change
- Add/edit a redirect rule
- Choose `http` or `https` in Redirect Protocol section
- Visit the route
- Check if it redirects to the protocol you've selected. 

### Changelog Entry

> Added - Custom redirect protocol per route

### Credits

Props @jayedul 

### Checklist:


Sidsector9 (changes requested):
@jayedul correct me if I'm wrong, the issue requests to implement a checkbox which enforces SSL rather than providing an option to choose the protocol.

jayedul:
@Sidsector9 Not to force, by default it will use current protocol. However if users want, they can change the protocol to either http/s for individual routes.

https://github.com/10up/safe-redirect-manager/issues/46#issue-21901904

jeffpaul:
@Sidsector9 bumping back to you for review

Sidsector9:
@jayedul after discussing this internally, we arrived at the conclusion that it would be better to only enforce https and remove the option to switch to http.

More details here - https://10up.slack.com/archives/C84M9AA06/p1684232341604279

jayedul:
@Sidsector9 Now radio button removed, and single checkbox added to enable force HTTPS redirect.
Screenshot 2023-06-01 at 2 45 30 PM

jeffpaul:
@Sidsector9 back to you for review here

Sidsector9 (approved):
Apologies for the delay in review. Thanks for the changes! LGTM 👍

PR #307, opened 2023-03-21
sorting status code array when additional are added

### LINKED ISSUE #308: Using the `srm_additional_status_codes` filter hook throws warning
### Describe the bug

When the `srm_additional_status_codes` filter hook is used to add more HTTP codes, it throws an "Undefined offset" PHP warning.

This is most likely caused due to the following line:

https://github.com/10up/safe-redirect-manager/blob/65b6913d1cb590a606dccecd98928e973ca3de74/inc/classes/class-srm-post-type.php#L33

...as the method `srm_get_valid_status_codes_data()` is directly called inside the constructor.
If you move this call inside the `init` hook, it works as expected.

We should investigate whether if going ahead with this change will break any existing functionality.

### Steps to Reproduce

Use the following snippet:

```php
add_filter( 'srm_additional_status_codes', function( $codes ) {
	$codes[304] = 'Go there!';

	return $codes
} );
```

Screenshot 2023-03-30 at 7 29 09 PM

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

norcross:
## Note

Merge PR #312 before this

### Description of the Change
This PR checks to see if any additional status codes have been added via the `srm_additional_status_codes` filter. If so, it creates the combined array and then sorts it by numeric key before returning it to the dropdown. This is a UX / UI issue, and does not impact any of the functionality already in the plugin.

As you can see below, when additional statuses are added, the list loses the numeric order and can be confusing.
dropdown-before

### How to test the Change

1. Add a new HTTP status code via filter
2. Confirm that the code is in numeric order

### Credits
Props @norcross

### Checklist:
- [ x ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ x ] I have updated the documentation accordingly.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -124,7 +124,15 @@ function srm_get_valid_status_codes_data() {
 		array()
 	);
 
-	return $status_codes + $additional_status_codes;
+	if ( empty( $additional_status_codes ) ) {
+		return $status_codes;
+	}
+
+	$status_code_array  = $status_codes + $additional_status_codes;
+
+	ksort( $status_code_array, SORT_NUMERIC );
+
+	return $status_code_array;
 }
 
 /**


jeffpaul:
Oh that's a really nice improvement there, thanks @norcross!

Sidsector9 (dismissed):
Thank you so much for the PR!
LGTM 👍

Sidsector9:
Related issue #308

Sidsector9 (approved):

PR #312, opened 2023-04-14
fix/308: Fix the initialisation order

### LINKED ISSUE #308: Using the `srm_additional_status_codes` filter hook throws warning

Sidsector9:
### Description of the Change

Closes #308 

### How to test the Change

This PR initialises the class properties on `init` hook  instead of directly inside constructor.

### Changelog Entry

> Fixed - Fix issue with `srm_additional_status_codes` filter hook

### Credits

Props @Sidsector9

### Checklist:


faisal-alvi (approved):
@Sidsector9 thanks for the PR. It is working as expected.

## develop branch

image

## fix branch

image

PR #313, opened 2023-05-02
Filter hooks docblocks added

### LINKED ISSUE #282: Document filters
### Is your enhancement related to a problem? Please describe.

Currently the plugin is lacking docblocks for the filters. Per the 10up best practices, it would be helpful to add these.

For example:
https://github.com/10up/safe-redirect-manager/blob/e5d018dc24e4bcc8092d2543ab05882a9a8d334a/inc/functions.php#L95-L103

### Designs

_No response_

### Describe alternatives you've considered

_No response_

### Code of Conduct

jayedul:
### Description of the Change
Dockblocks added for various filter hooks

Closes #282

### Changelog Entry

> Changed - Filter hooks dockblocks added

### Credits

Props @jayedul 

**Post merge to `trunk` action:**
We need to set the "gh-pages" branch in GitHub Pages to get this auto-deployed on docs site.

image

### Checklist:


inc/classes/class-srm-post-type.php
```diff
@@ -229,11 +229,18 @@ public function action_redirect_chain_alert() {
 		global $hook_suffix;
 		if ( $this->is_plugin_page() ) {
 
+			/**
+			 * Filter whether possible redirect loop checking is enabled or not
+			 *
+			 * @return bool
```
peterwilsoncc:
```suggestion
			 * Filter whether possible redirect loop checking is enabled or not.
			 *
			 * @param bool $check_possible_loop Whether to check for redirect loops. Default false.
```

inc/classes/class-srm-post-type.php
```diff
@@ -474,6 +482,11 @@ protected function get_redirect_capability() {
 			$role->add_cap( $redirect_capability );
 		}
 
+		/**
+		 * Filter the capability to manage redirects
+		 *
+		 * @return string
```
peterwilsoncc:
```suggestion
		 * Filter the capability required to manage redirects.
		 *
		 * @param string $redirect_capability The required capability. Default `srm_manage_redirects`.
```

inc/classes/class-srm-post-type.php
```diff
@@ -562,6 +575,11 @@ public function redirect_rule_metabox( $post ) {
 		$enable_regex   = get_post_meta( $post->ID, '_redirect_rule_from_regex', true );
 
 		if ( empty( $status_code ) ) {
+			/**
+			 * Filter the default status code to redirect with
+			 *
+			 * @return int
```
peterwilsoncc:
```suggestion
			 * Filter the default HTTP status code to redirect with.
			 *
			 * Which HTTP redirect code safe redirect manager should default to. This can
			 * be overridden in the dashboard for each redirect.
			 *
			 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
			 *
			 * @param int $default_status_code Default redirect statues. Default value `302`.
```

inc/classes/class-srm-redirect.php
```diff
@@ -38,8 +38,9 @@ public function setup_redirect() {
 		$this->multisite_checks();
 
 		/**
-		 * To only redirect on 404 pages, use:
-		 *   add_filter( 'srm_redirect_only_on_404', '__return_true' );
+		 * Filter whether to only redirect on 404 pages
+		 *
+		 * @return bool
```
peterwilsoncc:
```suggestion
		 * Filter whether to only redirect on 404 File Not Found pages.
		 *
		 * @param bool $404_redirects_only Whether to redirect file not found requests only. Default false.
```

inc/classes/class-srm-redirect.php
```diff
@@ -98,10 +99,18 @@ public function match_redirect( $requested_path ) {
 			$requested_path = '/';
 		}
 
-		// Allow redirects to be filtered
+		/**
+		 * Filter registered redirects
+		 *
+		 * @return array
```
peterwilsoncc:
```suggestion
		 * Filter registered redirects.
		 *
		 * Allows plugin developers to modify the available redirects.
		 *
		 * @param array $redirects Redirects found for the redirect path.
		 * @param string $requested_path Original path of the requested URL.
```

peterwilsoncc (commented):
WordPress hooks are usually documented with `@param` rather than the return value, for filters it is implied that the first parameter is the return value. See a [docblock from core for an example](https://github.com/WordPress/wordpress-develop/blob/c79107aed92452719892a61dba65ad25d14c1af3/src/wp-includes/cron.php#L736-L754).

I've added in some example based on the PHP standards.

10up tends to use https://github.com/matzeeable/wp-hookdoc for documenting hooks, although it hasn't been updated for a while. It allows us to create a github page with each of the hooks documented.

@dkotter do you have thoughts on whether to continue using wp-hookdoc and/or whether we should use the WP standard for the OSP plugins?

dkotter:
> do you have thoughts on whether to continue using wp-hookdoc and/or whether we should use the WP standard for the OSP plugins?

So we've continued to use `wp-hookdoc` but we have discussed switching to something else (and that is one of our OSP goals). We've talked about both https://github.com/WordPress/phpdoc-parser and https://github.com/wp-hooks/generator as possible replacements. I would love to switch to something that is more maintained and make that the new standard across all of our projects

peterwilsoncc:
Thanks Darin. I guess for the purpose of this PR it's best to use the existing standards so the team can work out a standard for the projects beforehand.

jayedul:
@Sidsector9 Your suggestions have been committed, and the remaining dockblocks updated using same pattern.

peterwilsoncc (changes requested):
Are you able to change these to use the https://github.com/matzeeable/wp-hookdoc standards, per https://github.com/10up/safe-redirect-manager/pull/313#issuecomment-1542408856

faisal-alvi:
@jayedul can you please check @peterwilsoncc's comment above?

faisal-alvi (dismissed):
@jayedul can you please resolve the conflicts too?

faisal-alvi:
@peterwilsoncc I've integrated https://github.com/matzeeable/wp-hookdoc and used it, please check.

iamdharmesh:
@peterwilsoncc @faisal-alvi I have added the deploy docs workflow and made some changes in hookdocs (added templates, styles, etc...).  This is ready for review now.

hookdoc-conf.json
peterwilsoncc, line 1: This will need to be added to the export ignore list it `.gitattributes`:

```
/hookdoc-conf.json export-ignore
```

Once committed you can test with the command `git archive HEAD -o ~/Downloads/srm.zip`

peterwilsoncc (commented):
Thanks, this looks great.

One note inline to avoid the config file ending up in the exported file/w.org repo.

hookdoc-conf.json
iamdharmesh: Thanks for catching this @peterwilsoncc. Handled in [52abf39](https://github.com/10up/safe-redirect-manager/pull/313/commits/52abf39acfa56d26e93c53e181a06871c05c86e5)

package.json
```diff
@@ -27,11 +27,14 @@
     "env": "wp-env",
     "env:start": "wp-env start",
     "env:stop": "wp-env stop",
-    "postenv:start": "./tests/bin/initialize.sh"
+    "postenv:start": "./tests/bin/initialize.sh",
+    "build:docs": "rm -rf out && jsdoc -c hookdoc-conf.json safe-redirect-manager.php inc"
```
peterwilsoncc:
```suggestion
    "build:docs": "rm -rf docs-built && jsdoc -c hookdoc-conf.json safe-redirect-manager.php inc"
```

peterwilsoncc (approved):
LGTM, caught a problem in the scripts which I've just fixed.

Thanks!

PR #318, opened 2023-05-31
fix cypress action

Sidsector9:
Hotfix for cypress action


iamdharmesh (approved):
Thanks for fixing this. LGTM 🚀

PR #340, opened 2023-08-18
Add a check to ensure we have a valid PHP version before loading plugin functionality and output an admin notice if needed

### LINKED ISSUE #335: Add PHP checks
More info: https://github.com/10up/convert-to-blocks/pull/123#issuecomment-1611475826

kmgalanakis:
### Description of the Change

In this PR, I'm adding a check for the minimum required PHP version before attempting to load the plugin files. If the condition is not met an admin notice is displayed and the plugin files are not loaded.

Closes https://github.com/10up/safe-redirect-manager/issues/335

### How to test the Change

* To see the admin notice and have the plugin functionality not loaded, try to load the plugin on a site with PHP version lower that 7.4.

### Changelog Entry

> Added - Check for minimum required PHP version before loading the plugin

### Credits

Props @kmgalanakis 

### Checklist:

### DIFF
--- a/safe-redirect-manager.php
+++ b/safe-redirect-manager.php
@@ -15,6 +15,55 @@
  * @package safe-redirect-manager
  */
 
+namespace SafeRedirectManager;
+
+/**
+ * Get the minimum version of PHP required by this plugin.
+ *
+ * @since 2.0.2
+ *
+ * @return string Minimum version required.
+ */
+function minimum_php_requirement(): string {
+	return '7.4';
+}
+
+/**
+ * Whether PHP installation meets the minimum requirements
+ *
+ * @since 2.0.2
+ *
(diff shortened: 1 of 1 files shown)


safe-redirect-manager.php
```diff
@@ -28,5 +77,5 @@
 	WP_CLI::add_command( 'safe-redirect-manager', 'SRM_WP_CLI' );
```
iamdharmesh:
```suggestion
	\WP_CLI::add_command( 'safe-redirect-manager', 'SRM_WP_CLI' );
```
Or maybe better to prefix added functions with `srm_`? As we are not using `namespace` throughout the plugin. 

Thanks

Sidsector9 (approved):

PR #341, opened 2023-08-27
fix/86: Display notice which vertices that cause a redirect cycle/loop 

### LINKED ISSUE #86: Simple loop detection
`check_for_possible_redirect_loops()` is kind of slow. We could add simple loop detection that runs in constant time by checking each redirect individually when it is created. It won't be as comprehensive, but will protect against the most blatant redirect loops.

We should block:
- [ ] any rule that redirects to itself
- [ ] any rule that redirects all traffic to the current site

Example of the second:

`/(.*)` -> `/$1/`

Redirecting ALL traffic to the same site will always be a loop.

### LINKED ISSUE #87: Improve check_for_possible_redirect_loops() efficiency
Currently, `check_for_possible_redirect_loops()` compares each redirect to every other redirect twice.

We should be able to do something like this to improve efficiency.

```
for ( $i = 0; $i < count( $redirects ); $i++ ) {
	for ( $j = $i + 1; $j < count( $redirects ); $j++ ) {
		// compare $redirects[$i] to $redirects[$j]
	}
}
```

If we're checking each redirect against itself when it is originally created, the inner loop can be `$j = $i + 1`, otherwise it needs to be `$j = $i`.

Sidsector9:
Closes #86
Closes #87

### Description of the Change

In PR #336, we use [DFS](https://en.wikipedia.org/wiki/Depth-first_search) to bring down the cycle detection time from `O(n^2)` to `O(V+E)`. We also returned as soon as the first cycle was detected.

In this PR, we make small adjustments to record all detected cycles by not returning early and then display the "Redirect To" values that cause cycles in a warning notice.

This PR also sets the default value of the filter `srm_check_for_possible_redirect_loops` to `true` as there is no negative performance hit and also helps to detect a redirect loop during the time of creating redirects.

Screenshot 2023-08-28 at 1 27 15 PM

### How to test the Change

1. Import the file  [srm-redirects-10000.csv](https://github.com/10up/safe-redirect-manager/files/12447775/srm-redirects-10000.csv). This file has 10,000 redirects.
2. Set the `srm_max_redirects` number to something more than 10,000.
3. Delete transients to gauge performance without it using `wp transient delete --all`
4. Visit `/wp-admin/edit.php?post_type=redirect_rule`

### Changelog Entry

> Changed - Include links to redirects that cause redirect loops/chains in the warning.
> Changed - Set the default value for the `srm_check_for_possible_redirect_loops` filter to `true`.

### Credits

Props @joshbetz @nicoladj77 @Sidsector9 

### Checklist:


ravinderk (changes requested):
@Sidsector9 Can we add conflicted redirects with their editable page link in admin notices? In this way, the admin can view the redirect URL edit page instead of searching for them.

Sidsector9:
@ravinderk in the implementation we have, we build an adjacency list which doesn't hold any post related info.

What you're requesting will require considerable refactoring. Best to handle it iteratively in a separate enhancement issue, what do you think?

ravinderk (approved):

PR #350, opened 2023-09-08
Quick Edit and Bulk edit

### LINKED ISSUE #98: Quick Edit and Bulk Edit
An idea for the plugin, maybe I'll create a PR in the coming days.

When you already have several redirections created but they are with the "wrong" HTTP Status Code, you have to manually to change them.

I suggest adding a quick edit and also a bulk edit to the plugin to make it easier to edit in bulk several redirections at once.

dhanendran:
### Description of the Change
Added option to Quick edit and Bulk edit redirect's HTTPS status and Force Https meta on the post list table. 

Closes #98 

Screenshot 2023-09-08 at 4 05 20 PM
Screenshot 2023-09-08 at 4 05 29 PM

### How to test the Change

1. Goto Safe Redirect Manager's list page
2. Hover over any redirect and confirm the "Quick Edit" option available
3. On the Quick Edit, confirm the Https status code and Force Https meta fields available to edit
4. Confirm the updates are saved to the post
5. Select multiple posts and select the "Edit" option from the bulk actions dropdown
6. Confirm the data updated to respective posts

### Credits
Props @dhanendran 

### Checklist:


inc/classes/class-srm-post-type.php
```diff
@@ -436,6 +424,56 @@ public function filter_redirect_sortable_columns( $columns ) {
...
+		}
+
+		if ( 'srm_redirect_rule_status_code' === $column_name ) :
+			wp_nonce_field( 'srm-save-redirect-ajax-meta', 'srm_redirect_ajax_nonce' );
+			?>
+			<fieldset class="inline-edit-col-right">
+				<div class="inline-edit-col">
+					<div class="inline-edit-group wp-clearfix">
+						<label class="inline-edit-status alignleft">
+							<span class="title">HTTP Status Code</span>
```
ravinderk: @dhanendran I noticed that most of the strings in this pull request are not translatable. Did you leave them intentionally?

assets/js/quick-bulk-edit.js
```diff
@@ -0,0 +1,28 @@
+( function ($) {
+	$( function () {
+		var wpInlineEditor = inlineEditPost.edit;
```
faisal-alvi: Consider using `let` or `const` instead of `var` including all locations below.

assets/js/quick-bulk-edit.js
```diff
@@ -0,0 +1,28 @@
...
+
+		inlineEditPost.edit = function (id) {
+
+			wpInlineEditor.apply(this, arguments);
+
+			var postId = 0;
+			if (typeof (id) == 'object') {
```
faisal-alvi: Use strict equality (`===`) instead of loose equality (`==`)

inc/classes/class-srm-post-type.php
```diff
@@ -436,6 +424,56 @@ public function filter_redirect_sortable_columns( $columns ) {
...
+							<select name="srm_redirect_rule_status_code">
+								<option value="-1">— No Change —</option>
```
faisal-alvi: Agree with @ravinderk, Please make sure we update all strings that need to be translatable.

faisal-alvi (changes requested):
@dhanendran thanks for the PR.

I have tested and the functionality is working as expected.

image

However, I just added some code changes below which should be addressed before we proceed with the merge.

inc/classes/class-srm-post-type.php
dhanendran: @ravinderk definitely not international. Now I have updated it properly.

dhanendran:
@faisal-alvi I have addressed your comments.

faisal-alvi (approved):

inc/classes/class-srm-post-type.php
```diff
@@ -746,6 +801,7 @@ public function load_resources() {
 		if ( 'redirect_rule' === get_post_type() ) {
 			wp_enqueue_style( 'redirectjs', plugin_dir_url( 'safe-redirect-manager/safe-redirect-manager.php' ) . 'assets/css/redirect.css', array(), SRM_VERSION );
 			wp_enqueue_script( 'redirectjs', plugin_dir_url( 'safe-redirect-manager/safe-redirect-manager.php' ) . 'assets/js/redirect.js', array( 'jquery' ), SRM_VERSION );
+			wp_enqueue_script( 'quick-bulk-editjs', plugin_dir_url( 'safe-redirect-manager/safe-redirect-manager.php' ) . 'assets/js/quick-bulk-edit.js', array( 'jquery', 'inline-edit-post' ), SRM_VERSION );
```
ravinderk: @dhanendran, I suggest not using a hard-coded plugin file name. It could cause an issue if the plugin directory name does not match.

cc: @faisal-alvi @dkotter

ravinderk (changes requested):
@dhanendran 

- [ ] When editing settings from quick edit options, It breaks the UI.
image

- [ ] Newly added setting fields are not aligned. You added an HTML class to align the setting fields to the right side, and each `fieldset` HTML tag adds a top margin, which looks bad.
image

assets/js/quick-bulk-edit.js
```diff
@@ -0,0 +1,28 @@
...
+
+			if (postId != 0) {
+				var editRow = $('#edit-' + postId);
+				var postRow = $('#post-' + postId);
+
+				var statusCode = $('.srm_redirect_rule_status_code', postRow).text();
+				var forceHttps = $('.srm_redirect_rule_force_https', postRow).text();
+
+				$('select[name="srm_redirect_rule_status_code"]', editRow).val(statusCode);
+				if ('✓' === forceHttps) {
```
ravinderk: @dhanendran I think we should not use symbols for comparison.


@dkotter do you have any thoughts?

dkotter: Yeah, this seems prone to breaking. Is there anything more specific we could target on, maybe we look for an empty string instead?

jeffpaul: @10up/open-source-practice curious if anyone else has insight on how to handle here?

faisal-alvi (changes requested):
@dhanendran thanks for the work on this.

1. I tested and found the bulk checkbox is not working.

350-bulk-check

2. Can you please also address the code review comments starting from https://github.com/10up/safe-redirect-manager/pull/350#discussion_r1327487260?

dhanendran:
@ravinderk @faisal-alvi I have addressed your comments.

@faisal-alvi The bulk select checkbox working fine for me. Can you test this again please?

assets/js/quick-bulk-edit.js
dhanendran: @ravinderk @dkotter @jeffpaul I'm using `x` character for comparison now.

faisal-alvi:
@dhanendran Thanks for the changes. I retested again in a different setup but the bulk select is not working there too. I am on WP v6.5.4, could you please try to replicate it in that version if you are not already on it?

350-gif-2

qasumitbagthariya:
### QA Update ❌
---

@dhanendran @faisal-alvi 
I've checked with a different setup but the bulk select is not working.

Jun-16-2024 15-22-05



**Testing Environment** 

<Details>

- WordPress: 6.5.4
- Theme: Twenty Twenty-Four 1.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS Ventura 13.3 

</Details>

safe-redirect-manager.php
```diff
@@ -71,6 +71,9 @@ function() {
 require_once dirname( __FILE__ ) . '/inc/classes/class-srm-loop-detection.php';
 
 define( 'SRM_VERSION', '2.1.2' );
+define( 'SRM_PLUGIN_FULL_FILE', __FILE__ );
+define( 'SRM_PLUGIN_URL', plugin_dir_url( SRM_PLUGIN_FULL_FILE ) );
+
```
faisal-alvi, line 76: This extra new line can be removed.

faisal-alvi (commented):
@mehul0810 Thanks for the fixes. The bulk select checkbox issue is resolved now. I have a minor "new line removal" suggestion below. Additionally, I'm curious about the necessity of custom JS code for [bulk edit](https://github.com/10up/safe-redirect-manager/pull/350/files#diff-822d37df0b2387357f3b4d66397a7e20d2fca8a2d7eee3c75260ab7fdd53bcec) and [bulk checkbox select](https://github.com/10up/safe-redirect-manager/pull/350/files#diff-68ab485c74cf32d4a09c8b9b240f3bc32ae4d3a09a6204c915de97635c8cc333) functionalities. Why doesn't the WordPress core handle these features by default?

Please also have a look at E2E failures.

mehul0810:
@faisal-alvi Removed the extra line and added a fix for the E2E tests.

> Additionally, I'm curious about the necessity of custom JS code for [bulk edit](https://github.com/10up/safe-redirect-manager/pull/350/files#diff-822d37df0b2387357f3b4d66397a7e20d2fca8a2d7eee3c75260ab7fdd53bcec) and [bulk checkbox select](https://github.com/10up/safe-redirect-manager/pull/350/files#diff-68ab485c74cf32d4a09c8b9b240f3bc32ae4d3a09a6204c915de97635c8cc333) functionalities. Why doesn't the WordPress core handle these features by default?

A custom Bulk Edit JS was already there and Bulk Select JS is something I have added after investigating the issue. We are not using an actual custom post type or WP_LIST_TABLE due to which the core functionality for bulk select or bulk edit was not working so we need to add those functionality. Let me know if you have any other questions. Thanks!

faisal-alvi (approved):

jeffpaul:
@faisal-alvi @dhanendran it would be great to generate a screenshot/screencast of this, add that to the dotorg directory, and then include in the screenshots section of the readme.txt file.

faisal-alvi:
@jeffpaul I haven't seen any screenshots added in the dotorg/readme yet. Should I include screenshots of the other plugin areas/pages as well? Also, do GIFs work on dotorg? A GIF might be better than a static screenshot for demonstrating bulk edit.

PR #369, opened 2024-02-24
Issue #366 - Make URL validation on input event instead of change event

### LINKED ISSUE #366: Redirect.js can leave publish button disabled when fields are valid
### Describe the bug

As the redirect.js file uses the `change` event to detect whether fields are valid, it requires users click out of the box before detecting the field has been filled out.

This can lead to confusion as the publish button remains disabled when the redirect fields have valid data, the user needs to click out of the field and then click the publish button.

This was reported in https://github.com/10up/safe-redirect-manager/issues/364#issuecomment-1939361111 (props @JosVelasco).

### Steps to Reproduce

1. Change the redirect status code to `301`
2. Click in to the redirect from to field
3. Type then delete some text
4. Click out of the redirect from field
5. The publish button should be disabled
6. Click in to the redirect from field
7. Type something, do NOT click out of the box
8. The publish button will remain disabled.

_No response_

### Environment information

- Device: Macbook
- Browser: Firefox 122/macOS 12.7.3 

### WordPress information

_No response_

### Code of Conduct

BhargavBhandari90:
### Description of the Change
Changed event for "Redirect From" field for URL validation.

- Previously it was on a `change` event, so validation was not happening unless we click out of the text field. Because of that publish button disabling was irregular and confusing.
- So made it on `keyup` event and added ajax abort to avoid simultaneous requests.
- Also added a typing wait time of 850 ms so when user stops typing, then URL validation will happen.
- Based on this validation process, disabling/enabling submit button became regular now.
- Also added a loding icon when user start typing, so that they know something is happening.

Video for better understanding for above bullet points https://somup.com/cZnTI6pzGz

Closes #366

### How to test the Change

- Go to WP admin --> Tools --> Safe Redirect Manager
- Click "Create Redirect Rule"
- Start typing into "Redirect From" and stop typing
- Remove some characters and stop
- Observe the submit button active/inactive state

### Credits
@BhargavBhandari90, @peterwilsoncc

### Checklist:

### DIFF
--- a/assets/js/redirect.js
+++ b/assets/js/redirect.js
@@ -3,26 +3,15 @@
 		const publishBtn = $('#publish');
 		const fromRule = $('#srm_redirect_rule_from');
 		const toRule = $('#srm_redirect_rule_to');
+		let timer = 0;
+		let currentRequest = null;
 
-		fromRule.change(function(el) {
+		fromRule.keyup(function(el) {
 			publishBtn.prop('disabled', true);
-			$.get(
-				window.ajaxurl,
-				{
-					action: 'srm_validate_from_url',
-					from: fromRule.val(),
-					_wpnonce: $('#srm_redirect_nonce').val()
-				}
-			).done(function( data ) {
-				if ( '1' === data ) {
-					$('#message').html( '' ).hide();
-					publishBtn.prop('disabled', false);
-				} else if ( '0' === data ) {
-					$('#message').html( `<p>${redirectValidation.urlError}</p>` ).show();
-				} else {
-					$('#message').html( `<p>${redirectValidation.fail.replace( '%s', data )}</p>` ).show();
-				}
-			});
+			if (timer) {
+				clearTimeout(timer); // Clear the time after function execution.
+			}
+			timer = setTimeout( validateUrl, 850); // Wait for 0.85 seconds.
 		});
 
 		// Show autocomplete for the 'Redirect To:' field.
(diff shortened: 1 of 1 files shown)


assets/js/redirect.js
```diff
@@ -3,26 +3,16 @@
...
-				if ( '1' === data ) {
-					$('#message').html( '' ).hide();
-					publishBtn.prop('disabled', false);
-				} else if ( '0' === data ) {
-					$('#message').html( `<p>${redirectValidation.urlError}</p>` ).show();
-				} else {
-					$('#message').html( `<p>${redirectValidation.fail.replace( '%s', data )}</p>` ).show();
-				}
-			});
+		fromRule.keyup(function(el) {
```
peterwilsoncc:
```suggestion
		fromRule.on('input', function(el) {
```

I think the [`input`](https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event) event is better here as it describes the exact use case: 

> The input event fires when the value of an [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input), [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), or [`<textarea>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) element has been changed as a direct result of a user action (such as typing in a textbox or checking a checkbox).

assets/js/redirect.js
```diff
@@ -96,5 +86,40 @@
...
+			currentRequest = $.ajax({
+				url: window.ajaxurl,
+				method : 'GET',
+				data : {
+					action: 'srm_validate_from_url',
+					from: fromRule.val(),
+					_wpnonce: $('#srm_redirect_nonce').val()
+				},
+				beforeSend : function() {
+					if ( currentRequest != null ) {
```
peterwilsoncc:
```suggestion
					if ( currentRequest !== null ) {
```

assets/js/redirect.js
```diff
@@ -96,5 +86,40 @@
...
+						currentRequest.abort();
+					}
+				},
+				success: function( data ) {
+
+					// Remove loader.
+					if ( fromRule.hasClass( 'ui-autocomplete-loading' ) ) {
+						fromRule.removeClass( 'ui-autocomplete-loading' );
+					}
```
peterwilsoncc:
```suggestion
					fromRule.removeClass( 'ui-autocomplete-loading' );
```

removeClass handles if the class doesn't exist so no need to check.

peterwilsoncc (commented):
I've added a couple of notes inline, thanks for the pull request.

assets/js/redirect.js
BhargavBhandari90: Thank you for this. I've made this change now.

assets/js/redirect.js
BhargavBhandari90: Done

assets/js/redirect.js
BhargavBhandari90: Thank you for this. I've made this change now.

BhargavBhandari90:
Thank you @peterwilsoncc for your suggestions.

jeffpaul:
@peterwilsoncc back to you for review/approval/merging

Sidsector9 (approved):
Tests well 👍

PR #404, opened 2024-11-12
Ensure the `add_new` and `add_new_item` labels are the same

dkotter:
### Description of the Change

In testing on WP 6.7, the Add New button for the Safe Redirect Manager post type is showing as `Safe Redirect Manager` instead of `Create Redirect Rule`. This causes an E2E failure and also doesn't make much sense, since this button shows right next to the title which is also Safe Redirect Manager.

Seems WP 6.7 changed something around the `add_new` and `add_new_item` labels so this PR ensures our post type uses the same text for both.

**Before on WP 6.7:**

Settings page before fix

**After on WP 6.7 (and older):**

Settings page after fix

### How to test the Change

1. Go to the Safe Redirect Manager page under Tools
2. Ensure the button shown at the top says `Create Redirect Rule`
3. Test on both WP 6.7 and 6.6

### Changelog Entry

> Fixed - Ensure the add new button shows proper text

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.

### DIFF
--- a/inc/classes/class-srm-post-type.php
+++ b/inc/classes/class-srm-post-type.php
@@ -602,8 +602,8 @@ class SRM_Post_Type {
 		$redirect_labels = array(
 			'name'               => esc_html_x( 'Safe Redirect Manager', 'post type general name', 'safe-redirect-manager' ),
 			'singular_name'      => esc_html_x( 'Redirect', 'post type singular name', 'safe-redirect-manager' ),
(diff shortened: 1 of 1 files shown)


jeffpaul (approved):
Looks good, waiting on e2e tests to pass before merging

PR #427, opened 2025-10-24
Update class-srm-redirect.php to check $_SERVER['REQUEST_URI']

turtlepod:
### Description of the Change

This PR fixes a PHP error reported by the WordPress coding standards:

> ❌ **Error (severity 5)**: Detected usage of a possibly undefined superglobal array index: `$_SERVER['REQUEST_URI']`.  
> Use `isset()` or `! empty()` to check if the index exists before using it.  
> *(WordPress.Security.ValidatedSanitizedInput.InputNotValidated)*

The issue was caused by incorrect logic when checking `$_SERVER['REQUEST_URI']`.  
This update ensures the value is validated before use, preventing potential PHP notices and improving code safety.

#### Benefits
- Prevents PHP “undefined index” warnings.
- Improves code reliability and security by ensuring proper input validation.
- Maintains compliance with WordPress coding standards.

---

### How to Test the Change

1. Test any functionality that relies on `$_SERVER['REQUEST_URI']` (e.g., URL-based logic or redirects).  
2. Verify that no PHP notices or warnings appear in the error logs.  
3. Confirm that all affected features continue to function as expected.

---

### Changelog Entry

**Fixed** – Added a proper check for `$_SERVER['REQUEST_URI']` to prevent undefined index errors and improve input validation.  
*(WordPress.Security.ValidatedSanitizedInput.InputNotValidated)*

---

### Credits

Props @turtlepod

---

### Checklist

- [x] I have tested the changes and confirmed no regressions.  
- [x] I have ensured compliance with WordPress coding standards.  


peterwilsoncc:
Coincidently I was looking at this a few minutes ago. WordPress ensures the value is populated in the `wp_fix_server_vars()`, so I think these are a false positive.

Let me know if I am missing something...

turtlepod:
@peterwilsoncc hmm.. interesting. 
but the initial code
```
sanitize_text_field( ... ) ?? ''
```
is a valid bug/incorrect code.

jeffpaul:
@peterwilsoncc note that @turtlepod has some OSS time this/next week, so if there's an update needed here please share

inc/classes/class-srm-redirect.php
```diff
@@ -294,7 +294,7 @@ public function maybe_redirect() {
 		 * @param {string} $request_path Request path. Default `$_SERVER['REQUEST_URI']`.
 		 * @returns {string} Request path.
 		 */
-		$requested_path   = esc_url_raw( apply_filters( 'srm_requested_path', sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ?? '' ) );
+		$requested_path   = esc_url_raw( apply_filters( 'srm_requested_path', sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ?? '' ) ) ) );
```
peterwilsoncc, line 296:
```suggestion
		$requested_path   = esc_url_raw( apply_filters( 'srm_requested_path', sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) );
```

I think the null coalescence can be removed all together as it's not possible for the value to be so in WP due to these lines.

https://github.com/WordPress/wordpress-develop/blob/f7bd33addcc31c568b51dbb4c04a37875f917d91/src/wp-includes/load.php#L37-L42

turtlepod: i don't think that's a good idea. even when it's not needed. it's still getting flagged by wp coding standard.

peterwilsoncc: Is the sniff because the global may not be defined, if so I think we can use a phpcs ignore rather than add code that won't be hit.

How about a phpcs:ignore WordPress.Sniff.Name -- see class-simple-local-avatars.php()

nhrrob:
Fix is already applied in the develop branch.
Nothing to merge.

Feel free to close this PR.

cc: @jeffpaul , @peterwilsoncc

PR #440, opened 2026-02-18
Update how we run PHPUnit

dkotter:
### Description of the Change

Our [PHPUnit workflow has started to fail](https://github.com/10up/safe-redirect-manager/actions/runs/22017529754) and in investigating that, seems GitHub has [updated their default version of Docker](https://github.com/actions/runner-images/issues/13474) that is installed in their runners and that is conflicting with the tool we use to run our tests. That tool hasn't been updated since June 2024 so I'm not confident they'll push out a fix.

Also, there's a better way to run PHPUnit tests now that didn't exist/we didn't know about when we set this up (using `wp-env`) so this is a good opportunity to clean up that setup.

The following updates are made:

- We update our GitHub Action workflow to use the `wp-env` environment to run our PHPUnit tests (and run those on a wider matrix)
- Update our various config and bootstrap files as needed
- Fixes a few PHP 8.4 issues

Note most of this was copied from the approach we took on [ClassifAI](https://github.com/10up/classifai/pull/1053).

### How to test the Change

Ensure the PHPUnit GitHub Action runs and passes on this PR

### Changelog Entry

> Fixed - PHP 8.4 deprecation warnings
> Developer - Update our PHPUnit workflow

### Credits

Props @dkotter

### Checklist:

- [x] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.


jeffpaul (dismissed):
Updated checks pass, looks fine to me

composer.json
```diff
@@ -10,7 +10,9 @@
   ],
   "require-dev": {
     "10up/phpcs-composer": "dev-master",
-    "yoast/phpunit-polyfills": "^1.0.3"
+    "phpunit/phpunit": "^8.5|^9.6",
```
dkotter, line 13: @peterwilsoncc I know I removed this on ClassifAI but in testing here, it was installing an older version of PHPUnit that relied on dependencies that weren't compatible with PHP 8.2+. Adding this line forced a newer version of PHPUnit to be installed that fixed that but let me know if there's a better approach here

peterwilsoncc (approved):

Issue #43, opened 2013-09-10
Add 'Export to .htaccess' feature


sxalexander:
Being able to setup and then export rewrite rules to an .htaccess format would be wonderful.

szepeviktor:
> Putting the rewite rules in your .htaccess can seriously affect your websites performance.

Not is you have access to the vhost configuration, those are parsed only once per Apache start.

PR #452, opened 2026-05-23
Add CSV/JSON export for redirects

nhrrob:
### Description of the Change

Adds the ability to export all redirect rules from the wp-admin redirect list
table as either a CSV or JSON file.

A format dropdown and an **Export** button appear in the top toolbar of the
Redirects screen. Selecting a format and clicking Export downloads the file
immediately. Each exported record includes: ID, Redirect From, Redirect To,
HTTP Status Code, Enable Regex, Force HTTPS, Message, and Status.

The export is protected by a nonce and a capability check using the existing
`srm_restrict_to_capability` filter, so only authorized users can trigger a
download. CSV output is also protected against formula injection attacks.

Closes #
Partially related to https://github.com/10up/safe-redirect-manager/issues/43 
After adding the Export feature, we can add the .htaccess export support too. PR is designed to be able to easily extend any future export types.

### How to test the Change

1. Go to the **Redirects** list table in wp-admin (Admin Dashboard -> Tools -> Safe Redirect Manager).
2. Confirm an **Export** feature (CSV / JSON dropdown and "Export" button) appear at the top of the table.
3. Select **CSV** → click **Export** → verify a `.csv` file downloads with all
   redirect data and correct column headers.
4. Select **JSON** → click **Export** → verify a `.json` file downloads as a
   valid JSON array with the same data.
5. Confirm the existing **Filter** button still works normally.
6. Log in as a user without redirect management permissions and confirm the
   Export button is not shown.

### Changelog Entry

> Added - Export redirect rules as CSV or JSON from the admin redirect list table.

### Credits

Props @nhrrob

### Checklist:

- [] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change. [Note: I see only 1 test file for the core feature. I tried to keep the PR as minimal as possible. If the code looks good and test cases are needed I will add the test cases. My initial thought was, big PR is difficult to get merged. Do suggest if I need to add tests. Thank you.]

Notes:
- Screenshot added.
- Screencast: https://drive.google.com/file/d/1spUyRB_u_UvAqloHdmalO23N90j_7RAe/view?usp=sharing 
- Considered the code for large data, as well as proper sanitization and tried to keep the change as minimal as possible.

export-csv-json-for-redirects


inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,258 @@
...
+		if ( ! current_user_can( apply_filters( 'srm_restrict_to_capability', 'srm_manage_redirects' ) ) ) {
+			wp_die( esc_html__( 'You do not have permission to export redirects.', 'safe-redirect-manager' ) );
+		}
+
+		// Validate format against whitelist; fall back to csv for unknown values.
+		if ( ! in_array( $export_format, $this->supported_formats, true ) ) {
+			$export_format = 'csv';
+		}
+
+		$redirects = srm_get_redirects( array( 'post_status' => 'any' ), true );
```
thrijith: This will stop at 1000 ( default limit )

https://github.com/10up/safe-redirect-manager/blob/7867d061cd4781a4f7df3db7cb950f5ec7b05949/inc/functions.php#L62-L65

Installs with >1000 redirects get a silently partial export, bad for a backup/migration feature.

Should it call the existing `srm_max_redirects_reached()` (`functions.php:97`) and surface a notice, or raise the cap for export?

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,258 @@
...
+	protected function normalize_redirect( $redirect ) {
+		return array(
+			'ID'            => $redirect['ID'],
+			'redirect_from' => $redirect['redirect_from'],
+			'redirect_to'   => $redirect['redirect_to'],
+			'status_code'   => (int) $redirect['status_code'],
+			'enable_regex'  => (bool) $redirect['enable_regex'],
+			'force_https'   => (bool) $redirect['force_https'],
+			'message'       => $redirect['message'],
+			'status'        => $redirect['post_status'],
```
thrijith: A CLI export already exists, `SRM_WP_CLI::export_csv`

https://github.com/10up/safe-redirect-manager/blob/7867d061cd4781a4f7df3db7cb950f5ec7b05949/inc/classes/class-srm-wp-cli.php#L316

CLI outputs 6 fields, this PR outputs 8 with different escaping. Should the field list + normalization be shared between CLI and admin (and both gain notes in export)?

Header inconsistency is **pre-existing**, not caused by this PR, but worth noting: three vocabularies already exist.

- CLI export (`redirect_from/status_code/enable_regex`).
- CLI import defaults (`source/target/code/regex/order/notes`)
- Admin export introduced in this PR (human labels `Redirect From`…). So no export round-trips into `import` without `--source=…` mapping flags.

If the export is meant for re-import, aligning on the importer's column names (and adding notes) would make it seamless.

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,258 @@
...
+			<select id="srm-export-format">
+				<?php foreach ( $this->supported_formats as $format ) : ?>
+					<option value="<?php echo esc_attr( $this->get_export_url( $format ) ); ?>">
+						<?php echo esc_html( $format_labels[ $format ] ?? strtoupper( $format ) ); ?>
+					</option>
+				<?php endforeach; ?>
+			</select>
+			<a href="<?php echo esc_url( $this->get_export_url( $this->supported_formats[0] ) ); ?>"
+				id="srm-export-btn" class="button">
+				<?php esc_html_e( 'Export', 'safe-redirect-manager' ); ?>
```
thrijith, line 91: No empty guard: neither the button nor `handle_export` checks the count, so zero redirects still shows the button and downloads a header-only CSV / `[]` JSON silently, whereas the CLI export *errors* on empty. 

Should we hide the button or notify for empty export action?

thrijith (changes requested):

nhrrob:
@thrijith 👋🏼 
Addressed all the feedbacks. Admin and CLI exports now share the same field list, both include notes, and both use shared CSV formula-injection protection. 

 I also kept the export cap override so admin export can include all redirects, with the export-specific transient cleaned up afterward.

inc/functions.php
```diff
@@ -88,6 +89,34 @@ function srm_get_redirects( $args = array(), $hard = false ) {
...
+/**
+ * Returns the shared list of fields used by the admin and WP-CLI exports.
+ *
+ * @since 2.2.3
+ * @return string[]
+ */
+function srm_get_export_fields() {
+	return apply_filters(
+		'srm_export_fields',
+		array( 'ID', 'redirect_from', 'redirect_to', 'status_code', 'enable_regex', 'post_status', 'notes' )
```
thrijith: Should we add message and force_https to fields?

inc/functions.php
```diff
@@ -88,6 +89,34 @@ function srm_get_redirects( $args = array(), $hard = false ) {
...
 
+/**
+ * Returns the shared list of fields used by the admin and WP-CLI exports.
+ *
+ * @since 2.2.3
+ * @return string[]
+ */
+function srm_get_export_fields() {
+	return apply_filters(
+		'srm_export_fields',
```
thrijith, line 120: Wondering if we should introduce new hook, export fields should contain all fields we have, if it isn't required for the user they will be able to remove columns from exported csv

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,232 @@
...
+
+		if ( ! current_user_can( apply_filters( 'srm_restrict_to_capability', 'srm_manage_redirects' ) ) ) {
+			wp_die( esc_html__( 'You do not have permission to export redirects.', 'safe-redirect-manager' ) );
+		}
+
+		if ( ! in_array( $export_format, $this->supported_formats, true ) ) {
+			$export_format = 'csv';
+		}
+
+		$export_limit = apply_filters( 'srm_export_max_redirects', PHP_INT_MAX );
```
thrijith: defaulting to `PHP_INT_MAX` will load **every** redirect into one in-memory array. On large sites that risks timeout.

Can we do this instead?

Stream in pages with a `do-while` instead of raising the cap. `srm_get_redirects` can't be reused for this, its internal `while` loop owns paging and `array_merge($defaults,$args)` lets a passed `paged` override the counter (re-queries the same page), and it always accumulates + caches. 

Can the row-builder be extracted https://github.com/10up/safe-redirect-manager/pull/452/changes#diff-f3a2689b39e6a1a9532d63ed0bc25e99cafc51ae5d7486c07b1e68e5a9681a3fR63-R71 into a helper (e.g. `srm_get_redirect_data($id)`) and run a paged `WP_Query` here in the export calling that helper.

thrijith:
> @thrijith 👋🏼 Addressed all the feedbacks. Admin and CLI exports now share the same field list, both include notes, and both use shared CSV formula-injection protection.
> 
> I also kept the export cap override so admin export can include all redirects, with the export-specific transient cleaned up afterward.

Thanks a lot for the great work and quick turn around @nhrrob, added a couple queries, apologies for the back and forth.

nhrrob:
@thrijith  Addressed the feedbacks along with few addition fixes (issues found during large(100k) data set tests):

- `srm_get_export_fields()` now includes all fields (added `message`, `force_https`), still filterable via `srm_export_fields`. New fields appended at the end to keep the CLI export's existing column order intact.
- Removed the `PHP_INT_MAX` cap/transient hack — export now streams redirects page-by-page (100/page) via `WP_Query` straight to output. Row-builder extracted into a shared `srm_get_redirect_data()` helper.
- Load-tested at 10k/50k/100k redirects and fixed two scale issues found along the way: unbounded object-cache growth (`wp_cache_flush_runtime()` per page) and N+1 queries from `fields => 'ids'` (`_prime_post_caches()`). Also dropped `SQL_CALC_FOUND_ROWS` (`no_found_rows`).
- Browser export at 100k hit a real 30s `max_execution_time` fatal — fixed with a guarded `set_time_limit()` (filterable via `srm_export_time_limit`), same pattern WP core uses for long-running admin ops.
- Added guards so `fopen()`/`wp_json_encode()` failures can't crash mid-download.

thrijith (commented):
PR approved.

The CLI export has changed as new fields have been added, something to note in changelog in addition to the new hooks that are introducted.

Passing to @jeffpaul for final review assignment

thrijith (dismissed):

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	/**
+	 * Handles the export download request, validates auth, and dispatches to the
+	 * appropriate format handler.
+	 *
+	 * @since 2.2.3
+	 * @return void
+	 */
+	public function handle_export() {
```
peterwilsoncc, line 124: Rather than firing on `admin_init` this should use a dedicated endpoint for the download. I'd suggest using an `admin-ajax.php` event to do so.

This will also reduce the risk of caches causing unexpected results.

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,308 @@
...
+		if ( ! $this->query_redirects_page( 1 )->have_posts() ) {
+			wp_die( esc_html__( 'There are no redirects to export.', 'safe-redirect-manager' ) );
+		}
+
+		while ( ob_get_level() ) {
+			ob_end_clean();
+		}
+
+		header( 'Pragma: no-cache' );
+		header( 'Expires: 0' );
```
peterwilsoncc: With the switch to admin-ajax, this won't be needed.

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,308 @@
...
+
+		$this->each_redirect_page(
+			function ( $redirect_ids ) use ( $handle ) {
+				foreach ( $redirect_ids as $redirect_id ) {
+					$redirect = srm_get_redirect_data( $redirect_id, true );
+					fputcsv( $handle, array_map( 'srm_escape_csv', $this->normalize_redirect( $redirect ) ), ',', '"', '\\' );
+				}
+			}
+		);
+		// phpcs:enable WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv
```
peterwilsoncc:
```suggestion
		// phpcs:enable
```

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	/**
+	 * Streams redirects as a JSON array to php://output, one page at a time.
+	 *
+	 * @since 2.2.3
+	 * @return void
+	 */
+	protected function export_json() {
+		echo "[\n";
```
peterwilsoncc: Instead of manually constructing JSON data, it's better to construct the array and json encode it at the end:

```php
// Pseudo code
$results = array();
foreach ( $redirect_ids as $redirect_id ) {
  $redirect = srm_get_redirect_data( $redirect_id, true );
  $results[] = $this->normalize_redirect( $redirect )
}

echo wp_json_encode( $results );
```

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,308 @@
...
+		while ( true ) {
+			$query = $this->query_redirects_page( $paged );
+
+			if ( ! $query->have_posts() ) {
+				break;
+			}
+
+			$callback( $query->posts );
+
+			wp_cache_flush_runtime();
```
peterwilsoncc: This is really heavy handed and will cause either the thrashing of in-memory cache or for sites without a persistent cache a number of repeated database calls.

Only the cache that is been targeted should be cleared.

What is it targeting?

peterwilsoncc (changes requested):
I've added a few notes inline.

The most important is the use of a dedicated endpoint rather than running the code on `admin_init`.

inc/classes/class-srm-export.php
nhrrob: It was targeting the post/postmeta cache _prime_post_caches() primes for each page's 100 IDs — just trying to keep memory flat across a 100k-row export. 

Now swapped it for:
wp_cache_delete_multiple( $query->posts, 'posts' );
wp_cache_delete_multiple( $query->posts, 'post_meta' );

nhrrob:
@peterwilsoncc 
Addressed all 5 points — switched to a dedicated wp_ajax_srm_export endpoint, dropped the now-unneeded buffer/cache-header handling, applied the phpcs:enable suggestion, scoped the cache cleanup to wp_cache_delete_multiple(), and switched JSON to build-then-encode-once.

inc/classes/class-srm-export.php
peterwilsoncc: @nhrrob I think it's fine to keep the data in the cache. There isn't a great deal of data and the site needs to query them on the front end so keeping them in the cache serves a purpose.

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,291 @@
...
+			wp_die( esc_html__( 'Security check failed.', 'safe-redirect-manager' ) );
+		}
+
+		if ( ! current_user_can( apply_filters( 'srm_restrict_to_capability', 'srm_manage_redirects' ) ) ) {
+			wp_die( esc_html__( 'You do not have permission to export redirects.', 'safe-redirect-manager' ) );
+		}
+
+		if ( ! in_array( $export_format, $this->supported_formats, true ) ) {
+			$export_format = 'csv';
+		}
```
peterwilsoncc: Avoids the single-use variable and the need for sniff suppression.

```suggestion
		if ( ! wp_verify_nonce( sanitize_key( wp_unslash( $_GET['_wpnonce'] ?? '' ) ), 'srm_export' ) ) {
			wp_die( esc_html__( 'Security check failed.', 'safe-redirect-manager' ) );
		}

		if ( ! current_user_can( apply_filters( 'srm_restrict_to_capability', 'srm_manage_redirects' ) ) ) {
			wp_die( esc_html__( 'You do not have permission to export redirects.', 'safe-redirect-manager' ) );
		}

		$export_format = sanitize_key( wp_unslash( $_GET['export_format'] ?? 'csv' ) );
		if ( ! in_array( $export_format, $this->supported_formats, true ) ) {
			$export_format = 'csv';
		}
```

inc/functions.php
```diff
@@ -64,21 +64,7 @@ function srm_get_redirects( $args = array(), $hard = false ) {
...
-					'enable_regex'  => (bool) get_post_meta( $redirect_id, '_redirect_rule_from_regex', true ),
-					'force_https'   => get_post_meta( $redirect_id, '_force_https', true ),
-				);
-
-				if ( $include_post_status_in_result ) {
-					$redirect_data['post_status'] = get_post_status( $redirect_id );
-				}
-
-				$redirects[] = $redirect_data;
+				$redirects[] = srm_get_redirect_data( $redirect_id, $include_post_status_in_result );
```
peterwilsoncc: This will add `notes` to the returned data & therefore include it in the transient when querying data on the front end. Please remove it from this part of the code as it's not required.

inc/classes/class-srm-export.php
```diff
@@ -0,0 +1,291 @@
...
+				break;
+			}
+
+			$callback( $query->posts );
+
+			wp_cache_delete_multiple( $query->posts, 'posts' );
+			wp_cache_delete_multiple( $query->posts, 'post_meta' );
+
+			++$paged;
+		}
```
peterwilsoncc: The maximum number of available redirects is limited by `srm_get_max_redirects()` so the export functionality will need to respect that. You'll need to do something in the export functions to ensure the maximum isn't exceeded when the posts per page isn't an exact division of `srm_get_max_redirects()`, for example someone sets it to `1250` via the filter.

I've removed the cache deletion as noted in my earlier comment, too.

```suggestion
		$paged = 1;
		$count = 0;

		while ( true ) {
			$query = $this->query_redirects_page( $paged );

			if ( ! $query->have_posts() ) {
				break;
			}

			$count += count( $query->posts );
			$callback( $query->posts );

			if ( $count >= srm_get_max_redirects() ) {
				break;
			}
			++$paged;
		}
```

peterwilsoncc (changes requested):
I've added a few more minor change requests inline below.

nhrrob:
@peterwilsoncc  
Here are the changes made as suggested:
- Shared logic — CLI and admin export now go through one path (srm_each_export_redirect() + srm_get_export_fields() + srm_escape_csv()), so output can't drift.
- Notes — now included in both exports (was blank in the CLI).
- Transient — notes never touch it. Export reads notes directly per-redirect; srm_get_redirects() never requests them.
- Max redirects — both cap at exactly srm_get_max_redirects() (per-row break, so 1250 yields 1250, not a page-boundary overshoot), matching srm_get_redirects().

peterwilsoncc (approved):
This looks good to me, thanks for your patience during the review process.

PR #464, opened 2026-09-10
Add a user capability check to the `srm_validate_from_url` method

dkotter:
### Description of the Change

Currently the `srm_validate_from_url` method does a nonce check before running, which is great. This nonce should only exist if a user already has the `srm_manage_redirects` but if this code is ever refactored in the future to run in a different context, there's a possibility that a user that doesn't have the right capability could trigger it.

So this PR adds in a `current_user_can` check mostly for future-proofing.

### How to test the Change

Right now, the `srm_validate_from_url` method only runs when an AJAX request is made to the `srm_validate_from_url` action. This AJAX request only fires when someone edits the `Redirect From` field when creating/editing a redirect rule.

1. Checkout this branch
2. Create a new redirect rule. Ensure you get no error message when entering a `Redirect From` value
3. Create a new redirect rule. Enter the same `Redirect From` value
4. Ensure you get an error message about an existing redirect with the same URL

### Changelog Entry

> Added - Run a `current_user_can` check when we validate a URL to ensure the user has proper permissions.

### Credits

Props @dkotter

### Checklist:

- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.


inc/classes/class-srm-post-type.php
```diff
@@ -967,6 +967,12 @@ public function srm_validate_from_url() {
 			die();
 		}
 
+		// Ensure the caller is allowed to make this request.
+		if ( ! current_user_can( $this->get_redirect_capability() ) ) {
```
peterwilsoncc, line 971: In `::srm_autocomplete()` the capability check is hard coded to `srm_manage_redirects`, should we change this too while tidying up permission checks?

Within `SRM_Export` the capability reuses the filter, so maybe that should use the `::get_redirect_capability()` method there too.

peterwilsoncc (commented):
This looks good to me but a question inline about being consistent with cap checks.

inc/classes/class-srm-post-type.php
dkotter: I've updated `srm_autocomplete` to use that in https://github.com/10up/safe-redirect-manager/pull/464/changes/22e2b3cf8398ba6d5018b5df12040147cf0f01ca.

As for the uses in `SRM_Export`, I agree it would be great to standardize on a single helper function/method here but right now, `get_redirect_capability` is a protected method so can't call it outside of the class. It also does two things:

1. Return the capability (passing it through a filter)
2. Add that capability to the administrator role

I think ideal would be to separate those two things and move the actual capability return to a new helper function (likely in `functions.php`). I can make that change just wondering if that's more than what we'd like to modify here

peterwilsoncc (approved):

